PR #3, opened 2017-12-06
Design Home Page

wgranger:
This PR contains quite a lot. This is the complete design of the landing page. There is also a separate section created for stats since I will likely need to have a function somewhere there to calculate the SVG pie chart. I'm open to any suggestions concerning organization.

Question/comment: The way routes are written changed a bit on the starter repo, and I haven't yet made any changes on this app (if we want it to operate similarly to ASM). Check how the router functions in the `App` component.


src/styles/components/home-page.styl
```diff
@@ -0,0 +1,189 @@
+@import "font-styles.styl"
```
shaunanoordin: I was going to ask "why do we need to have `@import "font-styles.styl` here?" but then I realised it's going to be something related to a ridiculous quirk about code precedence, blurgh. OK, I'll write a full comment about this.

shaunanoordin:
## Weird Quirks of Code
The following files...
- `styles/global/buttons.styl`
- `styles/components/home-page.styl`
- `styles/components/home-statistics.styl`
...all declare `@import "font-styles.styl"` so they can run the `@extend` Stylus command, e.g. `@extend .primary-label`

However, _only styles/global/buttons.styl_ actually needs to have that `@extend` command; the other two .styl sheets can do fine without it.

### WTF? Why?
Because the order of precedence, when it comes to importing .styl files.

In `styles/main.styl`, we see...
```
@import "global/*"
@import "components/*"
```
...which means the Stylus files will be imported in this (folder-wise, then alphabetical) order:
1. global/buttons.styl
2. global/font-styles.styl
3. components/home-page.styl
4. components/home-statistics.styl

Since the extended font-styles are already registered by Step 2, `home-page.styl` and `home-statistics.styl` already **implicitly** know what font-styles to extend. Conversely, since `buttons.styl` is _loaded_ before `font-styles.styl`, it has to **explicitly** use `@import` to know what font-styles are available.

### Why does this matter?
I don't trust implicit rules, like the order of imported Stylesheets being determined by their names. (i.e. imagine if `button.styl` didn't have that `@import`, and the only reason it worked was because font-styles were defined in `a-font-styles.styl`, and if someone _renamed a file_ the whole thing collapses.)

As for solutions... ngeeehhh.

One part of me wants to enforce an "always explicitly spell out your `@imports` so people know what your specific stylesheets depend on" rule and another part of me thinks simplicity is the answer, and that the "strict order of imports" should only be applied to the `global/` folder, and all other stylesheets in `components/` can assume the global font-styles/variables/etc are always available.

Any thoughts, @wgranger ?

Side note: Strangely, there's a weird caveat in the colour variables  (e.g. `$plum` ) in global/app-themes, since the folder precedence applies, but the alphabetical-files-first precedence doesn't. (So you can rename the file the `global/zzz-app-themes` and the `$plum` colour would still be correct for `global/buttons`, which would now have loaded before zzz-app-themes.)

PR #5, opened 2017-12-13
Design About Page

wgranger:
This PR designs the main about page based on designs from inVision https://projects.invisionapp.com/d/main#/console/10495701/254604074/preview


webpack.config.js
```diff
@@ -8,6 +8,7 @@ module.exports = {
   devtool: 'eval-source-map',
   devServer: {
     contentBase: path.join(__dirname, '/src/'),
+    historyApiFallback: true,
```
shaunanoordin: Curious, what is this for? I found the [documentation](https://webpack.github.io/docs/webpack-dev-server.html#the-historyapifallback-option) but I'm not sure what the application is.

src/components/about/Index.jsx
```diff
@@ -1,23 +1,107 @@
...
+            </div>
+
+          </div>
+        </div>
+        <div className="about-page__team">
+          <h2>The team</h2>
+          <div className="about-page__members">
+            {TheTeam.map((member) => {
+              return (
+                <div>
```
shaunanoordin: [minor] React wants these divs to have the `key` attribute.

shaunanoordin:
👍

webpack.config.js
wgranger: I was having errors when navigating the site, I would often get a sort of "cannot GET /about..." type error, and this was a solution I found when browsing around.

wgranger:
Ah yes, thanks for the sharp eye on that key issue.

webpack.config.js
shaunanoordin: Oh Webpack, you are a mysterious mistress full of mystery.

PR #6, opened 2017-12-14
Style Classifier Toolbars and Navbar

wgranger:
This PR begins styling some components on the classifier page as well as the navbar. Currently, none of the buttons have any functionality.


shaunanoordin:
Nyerp! We've got some merge conflicts, let me see if I can resolve them.

shaunanoordin:
OH! I just remembered, @wgranger there might be a way for us to make the Subject Viewer component go "mostly full-screen" (i.e. fill up the space of the full screen, but minus header and footer) if we use flex. Solution would look something like...

```
<App>
  <Header />
  <Content class="SubjectViewer" />
  <Footer />
</App>

App { width: 100vw; height: 100vh; overflow: auto; display: flex; flex-direction: column }  //Or maybe min-height: 100vh and standard overflow, to prevent weirdness with our popups & modals.
Header { flex: 0 0 auto }
Content { flex: 0 0 auto }
Content.SubjectViewer { flex: 1 1 auto }  //THIS IS THE KEY
Footer { flex: 0 0 auto }
```

The main issue with using flexbox, IIRC, is trying to figure out _when_ you can calculate the pixel dimensions of the Subject Viewer. Do you run `calculateSize(SubjectViewer)` on component init, on component mount, or some time later in the cycle? We'd need to wait until the flex-sizing takes place before we record the SubjectViewer's size for the purpose of calculating Subject scaling ratio, etc.

beckyrother:
Woo-hoo! Excited to see this coming together. @wgranger let's talk about what we can do with the header and footer to minimize the space they take up.

PR #7, opened 2017-12-15
Minor Reorganisation - Dynamic Config file, better Stylus, removed unused compos

shaunanoordin:
## PR Overview
This PR is a minor organisation of the code, with the only notable update being the config file.

- The `src/config.js` file now uses our standard dynamic config, which recognises different `?env=`ironments.
- Stylus files reorganised. Goddammit, as mentioned in PR #3 , I can't stand how the import precedence of the files in stylus/global is implicitly alphabetical. It's OK to assume all `global/app-theme.styl` and `global/font-styles.styl` variables are present by the time we hit `components/`, but the `global/` stylesheets _should_ explicitly import all required styles.
- Some compos were unused. Goodbye, `about/About.jsx` , so long, `about/NestedComponents`

### Status
Ready for review.


wgranger:
Thanks for doing this. It's nice to have the config and then we can fill in the necessary info as we go. I'm going to create an issue to get some project details set up (on production, staging, and doorkeeper) and go ahead and set up a test project.

PR #9, opened 2017-12-18
Introduce SVG Image to Classifier

wgranger:
This PR introduces a subject into the classifier to begin working with annotations and buttons.

12/18/2017 - Asking for an initial look over this branch to determine how subject should correctly be resized.

### DIFF
--- a/src/components/SVGImage.jsx
+++ b/src/components/SVGImage.jsx
@@ -1,0 +1,76 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+export default class SVGImage extends React.Component {
+  constructor(props) {
+    super(props);
+    this.state = {
+      loaded: false,
+      error: false,
+    };
+
+    this.image = new Image();
+    this.image.onload = () => {
+      if (this.props.onLoad) this.props.onLoad(this.image);
+      this.setState({
+        loaded: true,
+      });
+    };
+    this.image.onerror = (err) => {
+      if (this.props.onError) this.props.onError(err);
+      this.setState({
+        error: true,
+      });
+    };
+
+    if (this.props.src) {
+      this.image.src = this.props.src;
+    } else {
+      this.state.loaded = false;
+      this.state.error = true;
+    }
+  }
+
+  componentWillReceiveProps(nextProps) {
+    if (nextProps.src !== this.image.src) {
+      this.image.src = nextProps.src;
+    }
+  }
+
+  render() {
+    if (this.state.loaded) {
+      return (
+        <image className="svg-image"
+          xlinkHref={this.image.src}
+          width={this.image.width}
+          height={this.image.height}
+          x={(this.image.width * -0.5)+'px'}
+          y={(this.image.height * -0.5)+'px'} />
+      );
+
+    //TODO: review loading and error indicators.
+    } else if (this.state.error) {
+      return (
+        <g className="svg-image-error">
+          <path d="M -60 -80 L 0 -20 L 60 -80 L 80 -60 L 20 0 L 80 60 L 60 80 L 0 20 L -60 80 L -80 60 L -20 0 L -80 -60 Z" />
+        </g>
+      );
+    } else {
+      return (
+        <circle className="svg-image-loading" cx={0} cy={0} r={100} />
+      );
+    }
+  }
+}
+
+SVGImage.propTypes = {
+  src: PropTypes.string,
+  onLoad: PropTypes.func,
+  onError: PropTypes.func,
+};
+
+SVGImage.defaultProps = {
+  src: null,
+  onLoad: null,
+  onError: null,
+};
--- a/src/ducks/subject-viewer.js
+++ b/src/ducks/subject-viewer.js
@@ -1,0 +1,90 @@
+const MIN_SCALING = 0.1;
+const MAX_SCALING = 10;
+
+const initialState = {
+  frame: 0,
+  imageSize: { width: 0, height: 0 },
+  rotation: 0,
+  scaling: 1,
+  translationX: 0,
+  translationY: 0,
+  viewerSize: { width: 0, height: 0 },
+};
+
+const RESET_VIEW = 'RESET_VIEW';
+const UPDATE_IMAGE_SIZE = 'UPDATE_IMAGE_SIZE';
+const UPDATE_VIEWER_SIZE = 'UPDATE_VIEWER_SIZE';
+
+const subjectViewerReducer = (state = initialState, action) => {
+  switch (action.type) {
+    case RESET_VIEW:
+      let bestFitScale = 1;
+      if (state.viewerSize.width && state.viewerSize.height &&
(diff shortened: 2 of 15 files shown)


shaunanoordin:
## Quick Thoughts 1

The problem with the SubjectViewer's sizing (inherited from ASM) is due to these lines...
```
updateSize() {
    ...

    const ARBITRARY_OFFSET = 2;
    const w = this.section.offsetWidth - ARBITRARY_OFFSET;
    const h = this.section.offsetHeight - ARBITRARY_OFFSET;

    this.svg.setAttribute('viewBox', `${-w/2} ${(-h/2)} ${w} ${h}`);
    ...
  }
```

The logic encounters a chicken-and-egg problem. The `.svg` component tries to resize itself based on the size `.section` component. However, as per CSS logic, the size of the `.section` component depends on the size of its children... which includes the `.svg` component. Derp!

😵 

There are two solutions to this, and it involves either cooking either the chicken _OR_ the egg.

ONE: screw React, we'll resize the necessary sections using old school JS.

```
updateSize() {
  ...

  //THIS!!!
  const w = this.section.offsetWidth;
  const h = window.innerHeight - document.getElementsByClassName('app-header')[0].offsetHeight - document.getElementsByClassName('grommet')[0].offsetHeight;

  this.svg.setAttribute('viewBox', `${-w/2} ${(-h/2)} ${w} ${h}`);
  this.svg.style.width = w + 'px';
  this.svg.style.height = h + 'px';

  //And also, for illustrative purposes...
  document.getElementsByClassName('content-section')[0].style.height = h + 'px';
  document.getElementsByClassName('content-section')[0].style.overflow = 'hidden';

  ...
}
```

TWO: we re-examine how we use CSS sizing and SVG's weird sizing to get a proper fit. i.e. don't try to fight the visual layout using updateSize(). Maybe something like... svg.viewBox = a constant value, then use flexbox to make sure the SVG. I'll need to examine this option properly, though. ⭐️ 

ALSO NOTE: the current DOM structure of what's being resized means this is actually a chicken-and-egg-and-egg-and-egg problem, since if we look at this...

```
body
  div#root
    data[data-reactroot]
      header.app-header
      section.content-section  //This is the visual container (chicken)...
        section.project-header
        main.classifier  //...with lots of children components (eggs)...
          section.control-panel
          section  //...that determine its size...
            svg  //...but we only resize the SVG.
          section.toolbar
      div.grommet  //Side note: this footer needs a better name
```

shaunanoordin:
## Quick Thoughts 2
We can simplify the way we use our ducks.

Instead of writing...
```
//SubjectViewer.jsx

SubjectViewer.defaultProps = {
  currentSubject: null,
  frame: 0,
  rotation: 0,
  scaling: 1,
  translationX: 0,
  translationY: 0,
  user: null,
  viewerSize: {
    width: 0,
    height: 0,
  },
};
```

we can do...
```
//subject-viewer.js
const SV_INITIALSTATE = {
  svFrame: 0,
  svImageSize: { width: 0, height: 0 },
  svRotation: 0,
  svScaling: 1,
  svTranslationX: 0,
  svTranslationY: 0,
  svViewerSize: { width: 0, height: 0 },
};
export { SV_INITIALSTATE }

//SubjectViewer
SubjectViewer.defaultProps = {
  ...SV_INITIALSTATE
}
```

(Prefixes help avoid the confusion we encountered when trying to differentiate between, e.g., annotations.data vs previousAnnotations.data)

wgranger:
Ok, this gives me a good idea of where to begin. There might need to be a bit more adjusting to the section height, but it's good to know CSS plays a role there, and I recall a suggestion you made in an earlier PR about using Flexbox, which is probably relevant here. 

Also, React doesn't like the splat in default props, but there may be a way around this.

wgranger:
Alright. I have a solution that appears to work, but it differs slightly from the one mentioned above, mainly because subtracting the footer height made to subject image appear too small. Also, a major issue was having the scaling work correctly when the viewer size changed. 

Eventually, the control panel will need a media query when the screen height is a certain size, but I'll probably save that for a future PR.

src/ducks/subject-viewer.js
```diff
@@ -0,0 +1,102 @@
...
+        }
+      });
+
+    case UPDATE_VIEWER_SIZE:
+      let bestFitScaled = 1;
+      if (action.width && action.height &&
+          action.imageSize.width && action.imageSize.height) {
+        bestFitScaled = Math.min(
+          action.width / action.imageSize.width,
+          action.height / action.imageSize.height
```
wgranger: I'm aware this `bestScale` function should probably be abstracted instead of repeated (as it is used in `RESET_VIEW`. However, I've decided to hold off until adding the reset functionality.

src/components/SVGImage.jsx
```diff
@@ -0,0 +1,76 @@
...
+      return (
+        <image className="svg-image"
+          xlinkHref={this.image.src}
+          width={this.image.width}
+          height={this.image.height}
+          x={(this.image.width * -0.5)+'px'}
+          y={(this.image.height * -0.5)+'px'} />
+      );
+
+    //TODO: review loading and error indicators.
```
shaunanoordin: Can we check with @beckyrother on what's our Go-To symbol/sigil/icon/glyph/emoji/sacred mark for indicating "Something went terribly wrong" ? I think we already settled on a proper loading symbol too, didn't we?

None of the symbols I can think of works:
⚠️ - generic warning that makes errors seems more exciting that they should be; ❌  generic error/cancel symbol that can be confused for mathematical multiplication; ☠️  - incredibly confusing indicator that our website is either poisonous or stolen by pirates.

beckyrother:
@shaunanoordin where does this error appear? I'm most in favor of ⚠️  as it's familiar. `fa-warning` would be perfect.

The loading spinner is here: https://codepen.io/beckyrother/pen/VyaGEN (linked from #11 )

shaunanoordin:
@beckyrother It's _the "Subject Image" SVG itself_

screen shot 2017-12-20 at 18 29 23
_Currently what we see when the Subject Image encounters an error._

screen shot 2017-12-20 at 18 29 48
_Currently what we see when the Subject Image is loading._

The tricky part is that what's displaying the error is a HTML object where I can slot in fa-warning easily, but an SVG object. That said, I'm _very_ much in favour of standardising the icons/symbols, so let me figure out how to embed HTML objects/styles inside an SVG element. 👌

src/ducks/subject-viewer.js
shaunanoordin: 👍

wgranger:
There's also this screen if an image isn't available. https://projects.invisionapp.com/d/main#/console/10495701/270313708/preview
I'd like to add any other work (a spinner, didn't load image) for another PR to keep this one small, but I can always push up another commit if desired.

beckyrother:
Got it! Yes, please use that screen if the subject isn't available. Sorry for the confusion 😬

PR #12, opened 2017-12-21
Subject Loading and Error Messages

wgranger:
Based on conversation in #9, this PR adds functionality to the classifier by showing a loading message when a subject is being retrieved and a separate error message when a subject experiences a problem loading.


src/components/SubjectLoading.jsx
```diff
@@ -0,0 +1,52 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+const SubjectLoading = ({ loaded }) => {
+  const transform = !loaded ? 'scale(1)' : 'scale(5)';
```
shaunanoordin: Question: what's with the scaling here? Is it that, once the Subject Image has loaded, the "Subject Viewer Fit Image" action triggers and the Subject zooms out, and we need to compensate for that? If yes, it seems like we need another solution for it.

In any case, we should probably add a note here to clarify what's going on.

src/components/SubjectLoading.jsx
```diff
@@ -0,0 +1,52 @@
...
+            to="360 15 15"
+            dur="0.75s"
+            repeatCount="indefinite"
+          />
+        </g>
+      </g>
+      <text
+        x="0"
+        y="20"
+        fill="#4a4a4a"
```
shaunanoordin: Thing with SVGs is that I can't really decide on whether these styles should be separated into .styl files, or embedded as-is.

Don't change anything for this PR, but for future improvements, what are your thoughts?

src/containers/SubjectViewer.jsx
```diff
@@ -53,7 +56,7 @@ class SubjectViewer extends React.Component {
   }
 
   onImageLoad() {
-    if (this.svgImage.image) {
+    if (this.svgImage && this.svgImage.image) {
```
shaunanoordin: 👍

PR #13, opened 2017-12-21
Responsive Control Panel

wgranger:
This PR makes the right control panel more responsive on smaller screens. It also allows the panel to be hidden to allow more space for the subject viewer.

Design template can be found [here](https://projects.invisionapp.com/d/main#/console/10495701/264242741/preview).

### DIFF
--- a/src/components/ControlPanel.jsx
+++ b/src/components/ControlPanel.jsx
@@ -10,15 +10,13 @@ class ControlPanel extends React.Component {
     super(props);
 
     this.handleResize = this.handleResize.bind(this);
-    this.toggleInfo = this.toggleInfo.bind(this);
-    this.togglePanel = this.togglePanel.bind(this);
     this.toggleButton = this.toggleButton.bind(this);
     this.toggleIcon = this.toggleIcon.bind(this);
-    this.togglePeek = this.togglePeek.bind(this);
     this.toggleFieldGuide = this.toggleFieldGuide.bind(this);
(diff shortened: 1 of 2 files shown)


src/components/ControlPanel.jsx
```diff
@@ -25,28 +44,65 @@ class ControlPanel extends React.Component {
       <FieldGuide guide={this.props.guide} icons={this.props.icons} />, 'Field Guide'));
   }
 
-  render() {
-    const fieldGuideText = this.props.dialog ? 'Hide Field Guide' : 'Show Field Guide';
+  handleResize() {
+    if (window.innerHeight < 625 && this.state.showInfo) {
```
shaunanoordin: A part of me thinks that this hardcoded size-related value should be extracted out to a config file... a... `secondary-config-for-styles-i-guess.js` file, sorta. But another part of me just shrugs and thinks, this value is only used in this one component.

I'm going to OK this for now, but I have a feeling this may bite us in the far future.

wgranger: Yeah, it's hard to say. I can put it in a config file or at least make it a constant in the component if you think that'd help.

wgranger:
I shortened the panel on resize because it looked awkward overlapping the footer. Although the panel may still be too tall in some cases, this seemed to help somewhat, but maybe I was being too picky. This is worth reviewing again with @beckyrother once it is deployed.

PR #14, opened 2017-12-22
Button functions

wgranger:
This PR adds functionality to most of the buttons on the toolbar. Although this PR is rather large, the majority of the code comes directly from ASM without changes.

### DIFF
--- a/src/containers/SubjectViewer.jsx
+++ b/src/containers/SubjectViewer.jsx
@@ -4,12 +4,20 @@ import { connect } from 'react-redux';
 import { getSubjectLocation } from '../lib/get-subject-location';
 import { subjectError, SUBJECT_STATUS } from '../ducks/subject';
 import SubjectError from '../components/SubjectError';
+import { Utility } from '../lib/Utility';
 
 import {
-  resetView, updateImageSize, updateViewerSize
+  resetView, setTranslation,
+  updateImageSize,
+  updateViewerSize, SUBJECTVIEWER_STATE
 } from '../ducks/subject-viewer';
 
 import SVGImage from '../components/SVGImage';
+
+const INPUT_STATE = {
+  IDLE: 0,
+  ACTIVE: 1
+};
 
 class SubjectViewer extends React.Component {
   constructor(props) {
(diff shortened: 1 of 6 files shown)


src/containers/SubjectViewer.jsx
```diff
@@ -4,13 +4,21 @@ import { connect } from 'react-redux';
...
 import {
-  resetView, updateImageSize, updateViewerSize
+  resetView, setTranslation,
+  updateImageSize,
+  updateViewerSize, SUBJECTVIEWER_STATE
 } from '../ducks/subject-viewer';
 
 import SVGImage from '../components/SVGImage';
 
+const INPUT_STATE = {
```
shaunanoordin: This should eventually moved to a general constants lib, perhaps along with various other states. (e.g. FETCH_STATE to replace PROJECT_STATE, SUBJECT_STATE, etc since they're all the same IDLE/FETCHING/SUCCESS/FAILURE)

(But not replacing SUBJECTVIEWER_STATE or similar, which is unique to the specific duck.)

src/containers/SubjectViewer.jsx
```diff
@@ -25,9 +33,21 @@ class SubjectViewer extends React.Component {
     this.onImageLoad = this.onImageLoad.bind(this);
     this.onImageError = this.onImageError.bind(this);
     this.getBoundingBox = this.getBoundingBox.bind(this);
+    this.onMouseDown = this.onMouseDown.bind(this);
```
shaunanoordin: 👍 for the implementation as is, but someone please remind me...


1. I need to check if events can use React's onMouseDown and onMouseUp. I remember losing some data in the synthetic React events, but I should double check.
2. Also check if React's onMouseDown/Up/etc events also captures **touch input**
3. If either 1 or 2 doesn't pan out, I need to implement the touch input code. (Video games to the rescue!)

shaunanoordin: Re: item 1, Wait, what am I talking about? The `<svg>` element here DOES use proper React onMouseDown and onMouseUp. i.e. `<svg onMouseDown={this.onMouseDown}>`, now `this.svg.addEventHandler('onmousedown', this.onMouseDown)`

Sorry, it turns out my brain was on the fritz last night. OK, that just means I need to figure out if [React's Touch events.](https://reactjs.org/docs/events.html#touch-events)

src/ducks/subject-viewer.js
```diff
@@ -28,21 +40,55 @@ const subjectViewerReducer = (state = initialState, action) => {
...
+    case SET_SCALING: {
+      let newScale = (action.scale) ? action.scale : state.scaling;
+      newScale = Math.max(MIN_SCALING, Math.min(MAX_SCALING, newScale));
+
+      return Object.assign({}, state, {
+        scaling: newScale
+      });
+    }
+
+    case SET_TRANSLATION:
```
shaunanoordin: Out of curiosity, why are the switch-case statements now enclosed in `{}`? Usually I see this left out for simplicity -
 or is this the AirBnB style we're following? In any case, for consistency, `SET_TRANSLATION` is missing them brackets.

wgranger: The linter complains, but it is weird because it doesn't happen in every duck, and I couldn't figure out why linter was being so picky. I'm up for standardizing one way or another, but if we choose to remove the, we should update the linter so it won't complain. https://eslint.org/docs/rules/no-case-declarations

eatyourgreens: Here's how I've got switch/case formatting set up in PFE, from some code I was writing earlier.
```javascript
handleRadioKeyDown(questionId, answerId, e) {
    switch (e.which) {
      case BACKSPACE:
      case SPACE:
        if (e.target.checked) {
          e.preventDefault();
          this.handleRadioToggle(questionId, answerId, e);
        }
        break;
      default:
    }
  }
```

eatyourgreens: I think this is the corresponding line from `.eslintrc`.
`"indent": ["warn", 2, {"SwitchCase": 1}],`

PR #15, opened 2017-12-22
Show Tutorial

wgranger:
This PR allows the tutorial to appear. Note: there isn't currently a `workflowId` or tutorial set up for the production project. Also included in this PR was some more general setup on the ducks to provide resources needed to grab the tutorial.

In turn, an error was found in ASM which is documented in [this PR](https://github.com/zooniverse/anti-slavery-manuscripts/issues/200).

_**Note:**_ I noticed after opening this branch that there is a custom design to this tutorial visible [here](https://projects.invisionapp.com/d/main#/console/10495701/263176947/preview). I'd like to have a conversation about the best way to implement this design while using the ZRC component. I'm not sure it's possible to reproduce the design solely on creative CSS, but maybe there's another simple UI component that can be created in another PR while using the tutorial existence check functions in this branch.


shaunanoordin:
Conflict detected, resolving.

shaunanoordin:
Re: custom design for the tutorial, if we're going with a very specific tutorial design (they brownish-tan design that matches the rest of the website), then we might as well build our own little lighweight <Tutorial> component, cribbing as much code from the ZRC as possible.

Simply using the ZRC component and then adding custom overriding CSS is OK for minor tweaks, but at a glance, it looks like a very significantly different component.

src/components/App.jsx
```diff
@@ -16,6 +17,7 @@ import ProjectHeader from './ProjectHeader';
 class App extends React.Component {
   componentWillMount() {
     this.props.dispatch(fetchProject());
+    this.props.dispatch(fetchWorkflow());
```
shaunanoordin: My primary comment is that we need to keep an eye on this. Previously, we've hardcoded the workflow that our projects use within our config.js file. However, moving forward, the 'standard' Zooniverse practice is to set the Workflow ID within the Project settings, so we'll need to do something like... `fetchProject().then((proj)=>{ fetchWorkflow(proj.workflowid) })`

There are reasons why we _would want_ to use hardcoded Workflow IDs - e.g. with ASM, users in different Splits get different workflows - but they don't seem applicable to Scribes.

For now.

As far as I know.

Dun dun _dun._

But anyway, important thing is that we need to keep an eye on this. I'm aware that PR #18 makes a lot of changes to the API calls, so it's best that this PR isn't changed, and that any updates done in future PRs. (If 18 doesn't take care of this already.)

PR #16, opened 2017-12-26
Introduce Field Guide

wgranger:
This PR introduces the field guide into Scribes. There isn't currently a field guide attached to the production project, but the staging project should be set. This branch also creates a `Dialog` component that can be reused with other items.

### DIFF
--- a/src/components/ControlPanel.jsx
+++ b/src/components/ControlPanel.jsx
@@ -1,33 +1,74 @@
 import React from 'react';
+import PropTypes from 'prop-types';
+import { connect } from 'react-redux';
+import { fetchGuide } from '../ducks/field-guide';
+import { toggleDialog } from '../ducks/dialog';
+import FieldGuide from '../components/FieldGuide';
 
-export default function ControlPanel() {
-  return (
-    <section className="control-panel">
-      <div className="control-panel__header">
-        <h4 className="primary-label">Subject info</h4>
-        <span>&#x025B8;</span>
-      </div>
-      <hr />
-      <div className="control-panel__buttons">
-        <div>
-          <span className="primary-label">Name</span>
-          <span className="body-font">ENA NS 78 0117</span>
-        </div>
-        <div>
-          <span className="primary-label">Attribution</span>
-          <span className="body-font">Library of the Jewish Theological Seminary</span>
-        </div>
-        <a href="/" className="text-link">Collection Page</a>
-        <button className="button">Show Crib Sheet</button>
-        <button className="button">Show Page Reverse</button>
-        <button className="button">Show Field Guide</button>
-        <button className="button">Show Tutorial</button>
+class ControlPanel extends React.Component {
+  constructor(props) {
+    super(props);
 
-        <div>
-          <button className="button">Save Progress</button>
-          <button className="button button__dark">Finished</button>
+    this.toggleFieldGuide = this.toggleFieldGuide.bind(this);
+  }
+
+  componentWillMount() {
+    this.props.dispatch(fetchGuide());
+  }
+
+  toggleFieldGuide() {
+    this.props.dispatch(toggleDialog(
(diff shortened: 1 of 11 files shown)


src/components/Dialog.jsx
```diff
@@ -0,0 +1,85 @@
...
+}
+
+Dialog.defaultProps = {
+  dispatch: () => {},
+  size: { height: 200, width: 200 },
+  title: ''
+};
+
+Dialog.propTypes = {
+  children: PropTypes.node,
```
shaunanoordin: [minor] Do we need to set propTypes for children? Isn't that implicit from the fact it's React.Component already?

shaunanoordin:
Code LGTM at the moment as it's functional, but I'm wondering a few things:
- What is the purpose of the Dialog?
  - At the moment, it's a catch-all for a reposition-able mini window that supplements the main user tasks. (i.e. viewing the Field Guide, for now.)
  - Will we also use it as a **messaging** box, comparable to `alert()`? e.g. `alert("ERROR could not submit Classification")`
  - If not, what do we plan to use?
- Will we need the "block the whole screen" Popup (modal?) from ASM eventually?

src/components/ControlPanel.jsx
```diff
@@ -1,33 +1,85 @@
...
-          <span className="body-font">Library of the Jewish Theological Seminary</span>
+class ControlPanel extends React.Component {
+  constructor(props) {
+    super(props);
+
+    this.toggleFieldGuide = this.toggleFieldGuide.bind(this);
+  }
+
+  componentWillMount() {
+    this.props.dispatch(fetchGuide());
```
shaunanoordin: Nyuuurrr... after our experience with ASM, I'm starting to favour making all `fetch()` calls either
1. centralised in a main component (or root component), or better yet
2. centralised in a main lifecycle call (e.g. chain it in `fetchProject()` )

On the other hand, I can see the argument for keeping this as is: _"The Field Guide is only ever required when the Classifier Page is accessed, so there's no reason to load the FG when a user is simply looking at the site for its About page, dude."_

On the third flailing appendage, I can see a counter-argument to that argument, in which I say, _"Yeah but 1. this should be in the Classifier Page container not the Control Panel compo, 2. now you're making a call for the Field Guide EVERY time the Classifier Page loads which is a waste of resources, and 3. don't call me dude, man."_

src/components/Dialog.jsx
```diff
@@ -0,0 +1,85 @@
...
+        enableResizing={false}
+        minHeight={400}
+        minWidth={400}
+      >
+        <div
+          className="popup dialog"
+          tabIndex="0"
+          ref={(c) => { this.popupBody = c; }}
+          role="button"
+          onClick={(e) => { return e.target === this.popupBody && this.close(e); }}
```
shaunanoordin: WAIT! I forgot one thing!

Does this ever trigger? I think it was a carry over from the other screen-blocking Popup modal, where if you clicked the semi-transparent overlay, it was the equivalent of the Close action. In fact, I'm not sure we even need `this.popupBody`

src/components/ControlPanel.jsx
wgranger: No, I definitely want to have a sort of wrapper or root component that will collect all the resources the app needs. Should we have a sort of loader as in ASM or just return null until some sort of initialized prop is set? There are many ways to do this.

src/components/Dialog.jsx
wgranger: Gotcha. I can remove this in a future PR.

wgranger:
I think this Dialog component is mainly used for items which need to be dragged across the page (which will be necessary for the Hebrew/Arabic keyboards). For alerts, etc. I'd like to see if we can use Grommet's layer component, which acts as a modal.

PR #18, opened 2018-01-16
Fetch Core Resources Before Rendering

wgranger:
COOL! I think I did this right, and I'm pretty happy with how it has cleaned everything up. This PR abstracts all the core resources out to an initializer duck before rendering anything.

Let me know if something looks out of place here. It might be redundant to have an `initializerReady` prop as well as `INITIALIZER_STATUS` and also `login.initialized` (sp?) in another part of the code? Also, British "initialised" vs. American "initialized."

Also, some of these lines are general clean up with some of the ducks and removal of the unnecessary `popupbody` ref.

### DIFF
--- a/src/components/App.jsx
+++ b/src/components/App.jsx
@@ -3,8 +3,7 @@ import PropTypes from 'prop-types';
 import { connect } from 'react-redux';
 import { Route, Switch } from 'react-router-dom';
 import { ZooHeader, ZooFooter } from 'zooniverse-react-components';
-
-import { fetchProject } from '../ducks/project';
+import { fetchResources } from '../ducks/initialize';
 
 import AboutLayout from './about';
 import AuthContainer from '../containers/AuthContainer';
@@ -15,10 +14,12 @@ import ProjectHeader from './ProjectHeader';
 
 class App extends React.Component {
   componentWillMount() {
-    this.props.dispatch(fetchProject());
+    this.props.dispatch(fetchResources());
   }
 
   render() {
+    if (!this.props.initializerReady) return null;
+
     return (
       <div>
         <header className="app-header">
@@ -49,6 +50,7 @@ class App extends React.Component {
 
 App.propTypes = {
   dialog: PropTypes.node,
+  initializerReady: PropTypes.bool,
   location: PropTypes.shape({
     pathname: PropTypes.string
   })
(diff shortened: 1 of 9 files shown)


shaunanoordin:
Conflicts resolved, now let's test this out.

src/components/Dialog.jsx
```diff
@@ -39,9 +38,7 @@ class Dialog extends React.Component {
         <div
           className="popup dialog"
           tabIndex="0"
-          ref={(c) => { this.popupBody = c; }}
           role="button"
-          onClick={(e) => { return e.target === this.popupBody && this.close(e); }}
```
shaunanoordin: Reminder: this code was removed because the <Dialog> component doesn't have screen-spanning div that blocks interactions with elements behind it. This used to exist for a separate <Popup> _modal_ component, used in ASM.

src/ducks/login.js
```diff
@@ -14,15 +14,27 @@ const loginReducer = (state = initialState, action) => {
   switch (action.type) {
     case SET_LOGIN_USER:
       return {
-        user: action.user,  // null if logged out.
```
shaunanoordin: Why did we remove the comments?

src/ducks/initialize.js
```diff
@@ -0,0 +1,60 @@
...
+    default:
+      return state;
+  }
+};
+
+const fetchResources = () => {
+  return (dispatch) => {
+    dispatch({ type: FETCH_RESOURCES });
+
+    Promise.all([
```
shaunanoordin: Following my comments in PR #16, we need to keep an eye on `fetchWorkflow()`. If I understand correctly, Promise.all() will... run all the promises in parallel instead of in sequence, if I'm correct? (Someone check me here; I'll try to write a quick test to be sure.)

In any case, I don't see a guarantee that we'll get the Workflow ID from the Project resource _before_ we get the Project promise resolves; so to avoid using a hardcoded Workflow ID, we might still need to chain it like `fetchProject().then((proj)=>{ fetchWorkflow(proj.workflowid) })`

src/ducks/login.js
```diff
@@ -34,7 +46,6 @@ const checkLoginUser = () => {
 };
 
 const loginToPanoptes = () => {
-  // Returns a login page URL for the user to navigate to.
```
shaunanoordin: Comments, come back!

src/components/App.jsx
```diff
@@ -16,11 +14,12 @@ import ProjectHeader from './ProjectHeader';
 
 class App extends React.Component {
   componentWillMount() {
-    this.props.dispatch(fetchProject());
-    this.props.dispatch(fetchWorkflow());
+    this.props.dispatch(fetchResources());
   }
 
   render() {
+    if (!this.props.initializerReady) return null;
```
shaunanoordin: Reminder: add that loading symbol in a future PR.

PR #19, opened 2018-01-30
Create Annotation Tool

wgranger:
This PR could do a million more lines, but I've stopped it at this point to keep it from getting too large. Currently, this branch will create and store annotations, place those annotations in the Redux store, and open an annotation window. 

This branch does not save text or delete annotations. That functionality is best saved for a future PR to break things up. I would also like to combine a lot of the functionality between classifications and annotations (mainly updating the classification as annotations are stored and deleted), but I am going to wait on that work until after this is merged. 

There is also some information here such as storing some workflow info in a classification that returns `undefined` now but should be more functional once another current PR is merged.

### DIFF
--- a/package.json
+++ b/package.json
@@ -23,6 +23,7 @@
   },
   "homepage": "https://github.com/zooniverse/scribes-of-the-cairo-geniza",
   "dependencies": {
+    "counterpart": "^0.18.1",
     "history": "~4.6.3",
     "markdownz": "~7.3.1",
     "panoptes-client": "~2.8.0",
--- a/src/components/AnnotationsPane.jsx
+++ b/src/components/AnnotationsPane.jsx
@@ -4,10 +4,6 @@ import { connect } from 'react-redux';
 import { Utility } from '../lib/Utility';
 
 class AnnotationsPane extends React.Component {
-  constructor(props) {
-    super(props);
-  }
-
   renderAnnotationInProgress() {
     if (!this.props.annotationInProgress) return null;
 
@@ -33,7 +29,7 @@ class AnnotationsPane extends React.Component {
     );
   }
 
-  renderAnnotations(annotations) {
+  renderAnnotations() {
     if (!this.props.annotations.length) return null;
 
     const annotationPrefix = 'ANNOTATION_';
@@ -41,8 +37,8 @@ class AnnotationsPane extends React.Component {
     return this.props.annotations.map((annotation, index) => {
       if (annotation.frame !== this.props.frame) return null;
 
-      let onSelectAnnotation = this.props.onSelectAnnotation;
-      let svgPointPrefix = `ANNOTATION_${index}_POINT_`;
+      const onSelectAnnotation = this.props.onSelectAnnotation;
+      const svgPointPrefix = `ANNOTATION_${index}_POINT_`;
       const svgPoints = [];
 
       for (let i = 0; i < 2; i += 1) {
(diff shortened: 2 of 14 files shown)


src/components/AnnotationsPane.jsx
```diff
@@ -0,0 +1,115 @@
...
+
+    for (let i = 0; i < this.props.annotationInProgress.points.length; i += 1) {
+      const point = this.props.annotationInProgress.points[i];
+      svgPoints.push(
+        <circle
+          key={svgPointPrefix + i}
+          cx={point.x}
+          cy={point.y}
+          r={10}
+          fill="#20FECB"
```
shaunanoordin: Reminder: use constants instead of hardcoded style values. OK for a future cleanup PR.

src/containers/SubjectViewer.jsx
```diff
@@ -36,9 +41,15 @@ class SubjectViewer extends React.Component {
     this.onMouseDown = this.onMouseDown.bind(this);
     this.onMouseUp = this.onMouseUp.bind(this);
     this.onMouseMove = this.onMouseMove.bind(this);
-    this.onMouseLeave = this.onMouseLeave.bind(this);
```
shaunanoordin: Why remove onMouseLeave? With this gone, if a user click-drags to initiate navigation, and the mouse/cursor goes out of bounds (e.g. beyond the browser window), the app still thinks the user has the mouse button held down (i.e. the user is still in navigation mode) when the cursor re-enters the browser window.

wgranger: Ohhh. That's my bad. I forgot that functionality and was confusing it for `onMouseEnter`, which I believe was only added to ASM for the crop tool.

src/components/AnnotationsPane.jsx
wgranger: Are you referring to the hex code here? I can't get the variable from `app-theme.styl`, but I can put a constant in this component.

src/ducks/annotations.js
```diff
@@ -0,0 +1,72 @@
...
+      return Object.assign({}, state, {
+        status: ANNOTATION_STATUS.IDLE,
+        annotationInProgress: null,
+        annotations,
+        selectedAnnotation: state.annotationInProgress,
+        selectedAnnotationIndex: annotations.length - 1
+      });
+    }
+
+    default:
```
shaunanoordin: [super minor] Slight inconsistency with the use of `case thingy: { brackets }` here.

src/components/AnnotationsPane.jsx
shaunanoordin: 👍 Exactly what I was thinking. Maybe `./src/styles.js`

shaunanoordin:
👍 Nice, I see that onMouseLeave is back in!

package.json
```diff
@@ -23,6 +23,7 @@
   },
   "homepage": "https://github.com/zooniverse/scribes-of-the-cairo-geniza",
   "dependencies": {
+    "counterpart": "^0.18.1",
```
shaunanoordin: A thought: our eventual plan is that we'll implement the Translations system created by @eatyourgreens to handle Hebrew & Arabic labels.

When this happens, would we still need `counterpart`? I see that currently it's only used to determine the user's default language, as stored in Classifications.

eatyourgreens: The translations API will give you back the translated text for a Panoptes resource, like a tutorial or a workflow. You'll probably still need a translation library like counterpart to manage translations for the text in your interface, like the navigation links or button labels.

PR #21, opened 2018-02-05
Expand toolbar

wgranger:
This PR makes several changes based on [this design](https://projects.invisionapp.com/d/main#/console/10495701/257888022/preview). 

- Makes the right toolbar expandable
- Changes wording on subject info
- Adds background pattern to subject viewer
- Zooms the subject out a smidge on initial load to inform the user this is a subject to annotate (rather than a background)


shaunanoordin:
Merge conflicts resolved, but some quirks remain.

I added `&.active` to the...
```
button
  &:hover:not(.button-header), &:focus:not(.button-header) &.active
```
...CSS, but this doesn't really look that great, due to `border-radius: 50%` due to everything looking very rounded.

I can re-remove `&.active`, but this means we won't have a visual indicator when a button is selected (i.e. when choosing between Navigation or Annotation modes)

shaunanoordin:
Oh derp, that was a simple solution to `&.active`. There was a mirror to `border-radius` for `button.expanded`

PR #25, opened 2018-02-07
EXPERIMENTAL - Using a single Image file for an entire alphabet (Do NOT merge)

shaunanoordin:
## Experiment
Here's an experimental implementation - a way to compile all image assets representing an alphabet into a single `alphabet.png` instead of individual `a.png`, `b.png`, `c.png`, etc

screen shot 2018-02-07 at 17 09 10
_Example Arabic characters, as plastered on the home screen._

Pros:
- Easier asset maintenance - just use the CSS classes to represent a keyboard character.
- Fewer files for user to download

Cons:
- A lot of CSS calculations to be done, per character, albeit all upfront.
- A lot more calculations need to be done if you want components that scale to size.

### Status
For reference only. Do NOT merge.

@wgranger I hope this explains what I've been talking about. As mentioned, I do think a single image file-map has an advantage in maintenance, but _it's not a big deal_ since it's more a compression + maintenance trick. Having multiple image files, one per character, is a fine solution too, at the scale we're working with.


wgranger:
Ok. This makes sense and I see what you're talking about with the mapping. I'm still more inclined to use assets based on each image to avoid the CSS calculations (if I am handling the buttons/keyboard). That said, I think it becomes a matter of preference. If I end up handling this work, I'll opt for the asset per character, but if it turns out you cover the work on this keyboard, feel free to use the keyboard mapping shown here.

It does look like a really cool and clean method. I just hate falling down CSS styling holes and can see myself doing this and losing some time with this option.

wgranger:
Just keep in mind you'll also have a gridded view of the characters for each keyboard based on the InVision design.

srallen:
Image sprites help keep load times down. I used one with chimpandsee.

eatyourgreens:
A couple of questions, that might address the scaling problem with images:

Are the characters available in Unicode, or is this some ancient alphabet that no one uses any more?

Would it be too much effort to make a custom font for your Hebrew keyboard?

srallen:
I think font forge was used with plankton portal. There are probably even better ways of making custom fonts now. The custom font based on vectored images would be the ideal solution here I think.

wgranger:
@srallen If you experienced a noticeable improvement with load times, then that is a good reason to use the keyboard mapping approach here. 

@eatyourgreens These are all ancient texts with slight variations between then, so I think the Unicode option is out on that regard. I believe the team wanted an image of the character in script as an example (which they have provided), so I'm not sure creating the custom font based on vectored images would be a solution here, but I'll investigate Font Forge to confirm.

srallen:
@wgranger That's exactly what we used the custom font for in Plankton Portal. It's a font of the plankton examples used in the classifier and the field guide. You'll want vectored examples so it can render clearly on any screen.

wgranger:
Closing this as the approach has been implemented. Thanks for the help @shaunanoordin!

PR #27, opened 2018-02-08
Add Touch events to Subject Viewer

shaunanoordin:
## PR Overview
This PR attempts to add touch events to the Subject Viewer, allowing a user with an iPad (for example) to perform the same actions as a user wielding a computer with a mouse.

### Comments
Hey! I turns out I actually added touch sensing in `getPointerXY` in ASM, I just never hooked up the event listeners - so that's one major part of calculations already done. 👍 

```
SubjectViewer.getPointerXY(e) {
  const boundingBox = this.getBoundingBox();
  let clientX = 0;
  let clientY = 0;
  if (e.clientX && e.clientY) {
    clientX = e.clientX;
    clientY = e.clientY;
  } else if (e.touches && e.touches.length > 0 && e.touches[0].clientX &&  //Hahaha noice
      e.touches[0].clientY) {
    clientX = e.touches[0].clientX;
    clientY = e.touches[0].clientY;
  }
  ...
}
```

Adding `<svg onTouchStart={this.onMouseDown} etc etc >` seems to do most of the work, but I have an issue where touch-dragging (i.e. the equivalent of mouse-click-drag) _still causes the default "scroll webpage"_ action to take place. I suspect this has something to do with the Touch Event not being stopped properly, but debugging is still ongoing.

### Status
WIP

### DIFF
--- a/src/components/AnnotationsPane.jsx
+++ b/src/components/AnnotationsPane.jsx
@@ -65,9 +65,8 @@ class AnnotationsPane extends React.Component {
             }
             return Utility.stopEvent(e);
           }}
-          onMouseMove={e => Utility.stopEvent(e)}
           onMouseDown={e => Utility.stopEvent(e)}
-          onMouseUp={e => Utility.stopEvent(e)}
+          onTouchStart={e => Utility.stopEvent(e)}
         >
           {svgPoints}
         </g>
--- a/src/containers/SubjectViewer.jsx
+++ b/src/containers/SubjectViewer.jsx
@@ -254,6 +254,10 @@ class SubjectViewer extends React.Component {
           onMouseUp={this.onMouseUp}
           onMouseMove={this.onMouseMove}
           onMouseLeave={this.onMouseLeave}
+          onTouchStart={this.onMouseDown}
+          onTouchEnd={this.onMouseUp}
+          onTouchMove={this.onMouseMove}
+          onTouchCancel ={this.onMouseLeave}
         >
           <g transform={transform}>
             {subjectLocation && (
--- a/src/containers/SubjectViewer.jsx
+++ b/src/containers/SubjectViewer.jsx
@@ -119,6 +119,12 @@ class SubjectViewer extends React.Component {
         e.touches[0].clientY) {
       clientX = e.touches[0].clientX;
       clientY = e.touches[0].clientY;
+    } else if (e.changedTouches && e.changedTouches.length > 0 &&
+        e.changedTouches[0].clientX && e.changedTouches[0].clientY) {
+      //Special workaround for Chrome desktop in tablet emulation. Need to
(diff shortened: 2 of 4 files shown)


shaunanoordin:
## PR Update
VICTORY. When `onTouchStart={(e)=>{e.preventDefault();estopPropagation()}}` _doesn't_ prevent touch-dragging from causing you to scroll the page, the _CSS_ solution `touch-action: none` will do the trick.

https://caniuse.com/#search=touch-action

`.my-class { touch-action: none }` means that the `.my-class` element will ignore _default_ browser actions related to touch, touch-hold, touch-drag, etc. Strangely, this is mostly only applicable to Chrome, FF, and Edge, (the first of which I did touch testing on) and not Safari, so... I still need to test this on an iPad.

Let me break out the Xcode emulator _at least,_ but I'll need some help finding a free iPad.

### Status
Safe to start reviewing, but I'll also look at other onMouseDown events that might benefit from onTouchStart.

shaunanoordin:
Just a quick update: I've discovered a potential problem when testing with Chrome + OSX in tablet emulation mode. So, touch-dragging works fine in Navigation mode, but touch-tapping in Annotation mode is off; the start/end points are placed in a position that's completely incongruous to the position of the mouse.

It's a nice little puzzle to figure out, but I may need to finish this next week.

shaunanoordin:
Anyway, narrowed it down to this:
```
getPointerXY(e) {
    const boundingBox = this.getBoundingBox();
    let clientX = 0;
    let clientY = 0;
    if (e.clientX && e.clientY) {
      clientX = e.clientX;
      clientY = e.clientY;
    } else if (e.touches && e.touches.length > 0 && e.touches[0].clientX &&
        e.touches[0].clientY) {
      clientX = e.touches[0].clientX;
      clientY = e.touches[0].clientY;
    }
    ...
```

So on Chrome's tablet emulation, all variables the following variables don't register: `e.clientX, e.clientY, e.touches`. So it _should_ just be a matter of figuring out what the Synthetic React Event (tm)'s proper X-Y coordinates in that scenario should be.

Of course, I may just be overthinking this, and it turns out that the missing variables in Chrome's emulation mode actually work fine on the iPad. 🤷‍♂️

shaunanoordin:
## PR Update
AH! So `e.changedTouches` works. NOICE.

OK, the only thing left to do is test this on an iPad emulator and an actual iPad, and for that reason, I'm taking off the WIP here.

### Status
Ready for review.

wgranger:
Alright. The old `pointer-events` comes back. I remember having some issues with how this works a year ago in PFE, but it looks like you have a good grasp of it here and the code works well. LGTM.

wgranger:
I can confirm this is running on the Xcode (iPad) simulator as expected, but the same functionality is present (as expected) when running the site on a standard Mac.

PR #28, opened 2018-02-15
Begin Crib Sheet

wgranger:
First, I apologize for this PR being so large. I initially set out to reuse ASM code, but upon investigating the [design](https://projects.invisionapp.com/d/main#/console/10495701/278005669/preview) more, there appeared to be more functionality going on than first assumed. There are several components here along with the crop functionality.

Overall, this PR introduces the crib sheet to the site without the reference/keyboard functionality.


src/components/ActiveCard.jsx
```diff
@@ -0,0 +1,134 @@
...
+    } else if (emptyField && this.state.disableSave === false) {
+      this.setState({ disableSave: true });
+    }
+  }
+
+  editCardText() {
+    if (this.props.preferences.preferences && this.props.preferences.preferences.cribsheet) {
+      const cribCopy = this.props.preferences.preferences.cribsheet.slice();
+      cribCopy[this.props.cardIndex].name = this.inputText.value;
+      this.props.preferences.update({ 'preferences.cribsheet': cribCopy }).save();
```
shaunanoordin: I'm OK with `.save()` not providing success/failure feedback (since this is a non-core feature) but we need to be aware that any potential errors in this bit doesn't cascade out into the main app functinality.

shaunanoordin: To be fair, I am just paranoid after ASM.

src/components/ControlPanel.jsx
```diff
@@ -49,12 +51,19 @@ class ControlPanel extends React.Component {
   }
 
   toggleFieldGuide() {
-    if (this.props.dialog) {
+    if (this.props.dialogComponent === 'FieldGuide') {
```
shaunanoordin: I have to doublecheck the code to see what `dialogComponent` is for, but it might be worth turning these string into `const TYPE_FIELDGUIDE = 'FieldGuide';`, simply to avoid any potential muckups like `dialogComponent === 'fieldguide'` causing the code not to realise what you're talking about.

src/components/CribSheet.jsx
```diff
@@ -0,0 +1,198 @@
...
+
+      </div>
+    );
+  }
+
+  renderInstructions() {
+    return (
+      <div className="crib-sheet__personal-instructions handle">
+        <span>Use this crib sheet to save snippets for your personal reference.</span>
+        <span>If you&apos;re signed in, the images will be saved throughout your time on this project.</span>
```
shaunanoordin: Note a code issue, but a copy query: I'm not certain if "throughout your time on this project" communicates how this cribsheets actually work?

PR #29, opened 2018-02-16
IE11 Compatibility

shaunanoordin:
## PR Overview
Oh IE11, how could we have forgotten about you? (Oh, right: we _tried really hard.)_ This PR adds some quick fixes that allows IE11 users to access Scribes, and, just as importantly, allows local devs to test the website on their VMs because nobody actually installs IE11 on their machines if not held at gunpoint.

- Added polyfill service. Same as ASM, FYI. Added at the end <body> for performance reasons.
- Added a config option in webpack that disables the host-checking security mechanism; without this `disableHostCheck` option, accessing the localhost from a VM would lead to a fatal `Invalid Host header` error when accessing the webpage.

### Status
Simple fix, ready for review.

### DIFF
--- a/src/index.tpl.html
+++ b/src/index.tpl.html
@@ -15,5 +15,6 @@
   <body>
     <div id="root"></div>
     <%= htmlWebpackPlugin.options.gtm %>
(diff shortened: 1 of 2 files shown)


wgranger:
LGTM. Good catch on putting in the polyfills now rather than later.

PR #31, opened 2018-02-16
Begin Annotation Pane

wgranger:
This PR adds some functionality and styles to the annotation pane. The functionality should get a keen eye because, although they are similar to ASM, some functionality is a bit more sparse due to the lack of collaborative mode. Styles here do not yet include the keyboard.


shaunanoordin:
Yikes that is a lot of conflicts that need to be resolved. OK, gimme several seconds...

shaunanoordin:
Code looks good (with one exception, listed as a major thought below) and initial functionality tests are looking fine. 👍 

A few minor thoughts:
- Holy crap, the Annotation Box (aka transcription popup dialog aka SelectedAnnotation.jsx) is huge for this project - on a Macbook Pro, it covers the whole subject so you have to move it to see anything. Good thing you added draggable components!
- There's a tiny sliver (of brown) at the top of the Annotation Box where, if you click-drag, you select the text in the popup instead of dragging the popup around. Basically, there's one bit of the Dialog element that's not catching the click-drag event. Fairly minor, but somewhat annoying to users who might stumble across it, worth a quick fix later.

One major thought:
- Isn't this PR updating the Annotation BOX, not the Annotation Pane?

We already have a component specifically called AnnotationPane.jsx and that's the `SVG <g>` element that displays the start/end dots. It seems that this PR is updating the transcription dialog/popup, or Annotation Box, but definitely not the pane. The code _works,_ but for the use of, say, `annotationPanePosition` to denote the position of the SelectedAnnotation component will be confusing to future maintainers.

shaunanoordin:
Uh oh, I'm getting weird instances where if I do multiple annotations, then cancel out of them, then try annotating again, I have really weird unexpected behaviour. Let me see if I can dig more into this.

shaunanoordin:
OK, here's the issue

Issue: While in Transcription mode, after clicking on an existing Annotation and cancelling out, the Annotation state is all screwed up, and the next Annotation you make will "end" on the FIRST dot you place.

Steps:
1. Go to the Classify page.
2. Click on Transcription mode (in the toolbox on the left-hand side)
3. Make your first Annotation by marking the starting dot, the ending dot, and typing in "Hello world!" in the transcription box.
4. (Note that you are still in Transcription mode)
5. Click on the starting dot of your first Annotation. The transcription box (SelectedAnnotation) appears with "Hello world!"
6. Click Cancel to close the transcription box.
7. (Note that now you're in NEITHER transcription mode nor navigate mode on the Subject Viewer) ❗️ 
8. Click on Transcription mode (in the toolbox)
9. Make your second annotation, and...
10. ❗️Notice that the transcription box appears as soon as you lay down the starting dot.

WTF!

Analysis:
- There are two problems here.
- First, when Cancelling out of a selected existing annotation, the app doesn't know what was the previous state to revert to. (I'll need to look into what a good fix would be.)
- Second, if you pay close attention at Step 5, clicking on an existing annotation BOTH creates the start dot of a second annotation, AND opens up the transcription dialog.

I'm not sure what's happening with that second problem there, as the stopEvent() code should have prevented the click on the start/end dots (the Annotation) from bubbling up to the AnnotationPane. Eh, maybe one line of code is missing somewhere, should be fairly easy to trace - I'll look into it later.

wgranger:
I can add a delete button, and I'm thinking the delete button can take place of the cancel button? I recall we had a cancel and delete button in ASM because we had to deal with previous annotations and didn't want a user to delete a line of previous annotations.

That shouldn't be an issue in this project. If a user wants to remove the annotation box, they can just hit the X button on the top right of the dialog.

shaunanoordin:
👍 I'm OK with replacing the Cancel button with the Delete button (tagging @beckyrother just in case that's dumb) but don't forget that, functionally, we still need to account for 3 user actions: confirm (save changes & close popup), delete (delete annotation & close popup), cancel (don't make any changes & close popup) ... and each of these may require a special "what if" if the annotation text is blank. (If you cancel out of an annotation that had no text to begin with, should you delete the annotation too?)

shaunanoordin:
Full decision matrix should be something like...

| User Action | Condition | Result |
| -------- | -------- | -------- |
| Confirm | There's text | update annotation, close popup |
| Confirm | There's no text and this is a new annotation | delete annotation, close popup |
| Confirm | Text has been updated to blank and this is an existing annotation | delete annotation, close popup...? |
| Cancel | There's no text | delete the annotation, c. popup |
| Cancel | There's text | don't do anything to the annotation, c.p. |
| Delete | - | delete the annotation, c.p. |

beckyrother:
1. The X at the top of the dialog should only close the annotation box, NOT delete the annotation.

2. In that case, Cancel wouldn't be needed because the X would essentially do that action.

2a. If changes have been made and the X was clicked, need a confirmation box `Are you sure you want to close without saving? // No, continue transcribing / Yes, close without saving`

3. Ok to change the current Cancel button to Delete but it will also need a confirmation box `Are you sure you want to delete this transcription? // No, continue transcribing / Yes, delete`

src/components/ActiveCard.jsx
```diff
@@ -52,9 +54,32 @@ class ActiveCard extends React.Component {
     this.setState({ editBox });
   }
 
+  onDelete() {
+    if (this.props.preferences.preferences && this.props.preferences.preferences.cribsheet) {
+      const cribCopy = this.props.preferences.preferences.cribsheet.slice();
```
shaunanoordin: Is this safe or should this check for the existence of `this.props.preferences.preferences.cribsheet` before running `.slice()`?

wgranger: It appears to be checking the existence. Am I missing something?

src/components/QuestionPrompt.jsx
```diff
@@ -0,0 +1,47 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+const QuestionPrompt = (props) => {
+  return (
+    <div className="squestion-prompt" role="presentation">
```
shaunanoordin: Isn't this supposed to be `squestion-prompt`? The new stylesheet specifies `.question-prompt`, and this classname seems to break the style of the Question Prompt.

src/components/ActiveCard.jsx
shaunanoordin: Wait, derp, ignore my comment, I am obviously missing the obvious `if()` at the top there. Sorry, speed reading was a bad idea.

src/components/QuestionPrompt.jsx
wgranger: Oooo. Duh. Yeah, that's a typo I should've caught.

wgranger:
Ahh yes, I somehow missed this. `direction: rtl` does change a lot at the root level. However, some adjustments will still have to be made after this. Let's hold off on this merge for a moment while I take a look at that again. I may spin up another branch combining the two changes.

PR #34, opened 2018-03-05
[WIP] Switch Language Layouts

wgranger:
This PR does some preliminary work on switching language layouts (right to left). Although there is more work to do on this front, I think it makes sense to make small PRs since layout will have to be adjusted a bit with each component. Most importantly, this PR introduces a new duck which determines layout `languages.rtl`.

I think the best approach here moving forward is to use styled components, as this PR does in the second commit. It seemed to be the cleanest solution without getting into CSS hell.


eatyourgreens:
Why doesn’t this use CSS `direction:RTL` to flip the layout?

wgranger:
I've found with a good bit of the styling we have `direction: rtl` is insufficient, mainly because of using flexbox.

shaunanoordin:
## Further Comments
I'm normally in favour of keeping layout-related stuff contained to _centralised_ CSS/Stylus definitions. e.g. if `ducks.language.rtl === true`, then we'll have... `<div id="root" class="rtl">` and a bunch of classes in `layout.styl` reading:
```
#root .language-sensitive-flex-container { flex-direction: row; }
#root.rtl .language-sensitive-flex-container { flex-direction: row-reverse; }
```

That said, @wgranger , your implementation of Styled Components looks really good, so I'm all 👌  
 for these 'inline/in-JS-code styles'. I'm also aware that Sarah supports the usage of Styled Components, so I'm always happy to see the tech experimented with on projects outside of the Edu API FE. 👍 

My only caveat is that we ensure the inline styles doesn't make the code look too messy: the wild card in this scenario is what happens when we start adding text translations to individual components. (At that point, does the inline flex styling also have to compete with language RtL coding, etc? I have no idea.)

EDIT: @eatyourgreens I just saw your comments about `direction: rtl`. I haven't played with that myself, so let me read up on that.

wgranger:
But I could be missing something, and I need to read up on more on use cases for when exactly to use `direction: rtl`

shaunanoordin:
Update: looks legit.

@eatyourgreens if I understand @wgranger this PR is mostly focussed on having the _layout_ change from LtR to RtL, depending on the language. The actual _text direction_ (which I understand is what CSS `direction: rtl` will change) will be implemented in a later update/separate PR when we start having Hebrew translations.

For this PR, the intent of the Styled Components + this.props.rtl + flex-direction is to switch between these two views:

screen shot 2018-03-05 at 18 59 13
"English"

screen shot 2018-03-05 at 18 59 23
"Hebrew"

The text direction remains the same, for now.

eatyourgreens:
`direction: rtl` will flip flexbox layouts for you. Floated layouts have to be changed manually. Have a look at https://rtl.pfe-preview.zooniverse.org and notice how changing direction on the root element changes the layout of the classifier and also the navigation menus.

shaunanoordin:
@srallen Thanks for the documentation, although I have to admit I still can't get used to how the props are declared in a template literal.

Every time I see the pattern in a review, I go _what sorcery is this?? -oh right, that's how styled components roll._ then the next, _what sorcery is this?! right, that's how style components roll._ and then, _what sorcery is this!! oh right-_ etc

srallen:
Here's an example from our experimental use from PFE:

In this component, there's a defined wrapper, `MenuWrapper` using styled-components. A height prop is being passed to it in the render:

https://github.com/zooniverse/Panoptes-Front-End/blob/9c25f850e6aa67717e60f1c861ba367bbf39b909/app/pages/project/components/ProjectNavbar/components/ProjectNavbarNarrow/NarrowMenu.jsx#L58

The styled component, `MenuWrapper` is defined:

https://github.com/zooniverse/Panoptes-Front-End/blob/9c25f850e6aa67717e60f1c861ba367bbf39b909/app/pages/project/components/ProjectNavbar/components/ProjectNavbarNarrow/NarrowMenu.jsx#L10

According to the styled-component docs, this is a bit of syntactic sugar and shorthand for a factory:
https://www.styled-components.com/docs/api#styled

`styled.div` is equivalent to `styled('div')`, but the shorthand is preferred according to the docs.

The factory will use the template literal and the props to generate the stylesheet, inject the styles, and create the React component with the generated class names.

So in the example from PFE, you can see a function being called in the template literal with string interpolation:

https://github.com/zooniverse/Panoptes-Front-End/blob/9c25f850e6aa67717e60f1c861ba367bbf39b909/app/pages/project/components/ProjectNavbar/components/ProjectNavbarNarrow/NarrowMenu.jsx#L20

This is evaluated when the javascript is run and then used in the styled-components factory for the end result.

srallen:
There are a lot of ternary expressed that are being indirectly used, so for instance, this can be refactored:
```
const rtlSwitch = this.props.rtl ? 'row-reverse' : 'row';
const Button = styled.button`
  flex-direction: ${rtlSwitch};
`;
```

To instead do this:
```
const Button = styled.button`
  flex-direction: ${(props) => props.rtl ? 'row-reverse' ? 'row' }
`;
```
and the render passes that prop directly:
```
<Button rtl={this.props.rtl} />
```

wgranger:
Ahh yes, I somehow missed this. direction: rtl does change a lot at the root level. However, some adjustments will still have to be made after this. Let's hold off on this merge for a moment while I take a look at that again. I may spin up another branch combining the two changes.

eatyourgreens:
Also, I’d recommend adding a `lang` attribute in your markup, to declare the language to the browser(eg. To select the voice that VoiceOver will use to read the content.) You can use `[lang=he]` as your CSS selector to switch the direction for Hebrew at a fairly high level, then just override in your component styles where needed. 

If you switch galaxyzoo.org to Hebrew, and inspect the styles, then that’s an example of this technique.

wgranger:
Nice. I think that was probably the right approach to being with, but I failed to notice how effective it was. I'm playing with that a bit on another branch, and it is proving useful.

wgranger:
Closing in favor of #36

PR #35, opened 2018-03-07
Add Keyboards to Annotation Box

wgranger:
This PR adds necessary keyboards to the annotation box. This PR does not yet handle text input for the characters, but simply allows the box to switch through different views for the different keyboards.


src/components/SelectedAnnotation.jsx
```diff
@@ -23,10 +26,16 @@ class SelectedAnnotation extends React.Component {
...
+    this.setModern = this.setModern.bind(this);
+    this.toggleScriptOptions = this.toggleScriptOptions.bind(this);
+    this.closeDropdown = this.closeDropdown.bind(this);
+    this.previousScript = this.previousScript.bind(this);
+    this.nextScript = this.nextScript.bind(this);
 
     this.state = {
       annotationText: '',
-      showKeyboard: true
+      showKeyboard: true,
```
shaunanoordin: `showKeyboard` is in the state instead of the duck; this means that every time a user starts transcribing, the onscreen keyboard will always be open. (As opposed to, e.g. the `activeScript` variable, whose state is persistent.)

Wouldn't it be a better idea to move this to the duck so it remembers the user's choice?

src/components/SelectedAnnotation.jsx
```diff
@@ -179,24 +244,31 @@ class SelectedAnnotation extends React.Component {
             <hr />
             <span className="secondary-label">Current Script Type</span>
             <div>
-              <div>
-                <span>&#9664;</span>
-                <span className="text-link">Current Keyboard</span>
-                <span>&#9658;</span>
+              <div className="selected-annotation__script-select">
+                <button onClick={this.previousScript}>&#9664;</button>
+                <button className="text-link" onClick={this.toggleScriptOptions}>{this.props.activeScript.name}</button>
```
shaunanoordin: [minor][note] This doesn't check for the existence of `this.props.activeScript`, but a quick check of the code logic indicates that it shouldn't be possible for `this.props.activeScript` to be null or undefined.

src/lib/KeyboardTypes.js
```diff
@@ -0,0 +1,110 @@
...
+  CURSIVE: 'cursive',
+  MINISCULE: 'miniscule',
+  SQUARE: 'square'
+};
+
+const KeyboardOptions = [
+  {
+    name: 'Ashkenazi Cursive',
+    type: KEYBOARD_TYPES.CURSIVE,
+    img: '../images/keyboards/ashkenazi-cursive.jpg'
```
shaunanoordin: Did I mention how I really, really like how you're organising the script characters into a consolidated image file instead of having one .jpg file _per character?_ will++, becky++

wgranger:
A quick review of a call Shaun and I had.

There are a few changes which need to be made on this branch that will be done in a future PR. Namely: saving the open/close keyboard state and disabling script change with modern characters enabled.

Comments: 
- The modern keyboard should be active by default
- There is still some text direction work to be done on the input box which will be sorted out when text input is working with Hebrew characters.

PR #36, opened 2018-03-09
Second Go at Language Flip

wgranger:
This is a second go at flipping the screen based on selection of Arabic, Hebrew, or English text. I also committed the package.lock with this branch, but I can remove it if we decide to for any reason.

This proved to be a more succinct way of writing code with the correct implementation of direction: rtl and also while using styled components where needed.


shaunanoordin:
Two things:
1. there's a minor conflict that I'm going to resolve manually, no big.
2. something's wrong with two elements - "Show Previous Marks" and "Show Modern Characters" doesn't flip properly in the RtL layout.

screen shot 2018-03-09 at 18 44 53

wgranger:
Ah yes, a note on that. There are still some elements I haven't checked on the flip, so if you actually want to wait on merge, that is totally fine.

shaunanoordin:
Whoah, I just went through the CSS to take a better look at what's happening and I'm only just realising that I didn't properly check on how the components were styled.

Of note:
- The circle element for `.round-toggle` is actually a <label> that's shrunk to tiny proportions, and the text I see is actually overflowing from the circle. ❗️ 
- The `.round-toggle` that's used by SelectedAnnotation.jsx actually has CSS that's sitting in `collections-manager.styl` ❓ 

The `.round-toggle` looks like a jury-rigged solution for a custom `<input type="checkbox">` , and while I do see a few [articles online](https://www.inserthtml.com/2012/06/custom-form-radio-checkbox/) discussing this implementation as a cross-browser solution, it looks unnecessarily convoluted to me. (i.e. it took me a while to figure out what's happening, which isn't usually a good sign for future maintenance.)

I'm not sure what to say about this; I don't like that it's complex, but I can't argue that it's _functional._ 

My personal preference is to create a React component like `<RoundToggle label="Show Previous Marks" checked="true" direction="rtl" />` that handles the presentation within its state logic (and using an SVG for the checkbox, also to account for keyboard navigation), but that may just be moving the convoluted logic from CSS to JS.

If we're sticking with the current CSS-focused implementation, the solution is as follows:
```
//In style/components/collections-manager (!!!)

.round-toggle
  ...

  label
    ...
    //Remove left: 0  - this is unnecessary
    ...

    ...

  span
    ...
    //Remove margin-left: 1rem - this only accommodates LtR layouts
    margin: 0 1rem  //Add left and right margins
```

Although it's worth checking any other component using .round-toggle

PR #38, opened 2018-03-12
Add Script References

wgranger:
Although there are many lines touched in this PR, a good bit of it is setup for the keyboard (including CSS for the sprites). Also, open to suggestions on some of the functionality of the ScriptReferences component.


shaunanoordin:
Quick thoughts
- After clicking "Send to Keyboard" after selecting a script, there is no confirmation what happened.
  - Recommendation: at least have a popup that says "You will see this keyboard the next time you make an annotation". @beckyrother any better suggestions?
- If Modern Characters is selected, then when you click "Send to Keyboard", the next time you transcribe you will see Modern Characters instead of the script style you selected
  - Recommendation: when triggering "Send to Keyboard", the app needs to 1. set showKeyboard = true, and 2. set modernCharacters = false, _in addition_ to setting keyboardStyle = selectedKeyboard.
- When a **first time user** goes to Crib Sheet -> Your Crib Sheet, the component crashes.
  - This is fixed on a page reload...
  - ...and is probably unrelated directly to this PR's additions. I'll look into including a fix here, though.

screen shot 2018-03-13 at 16 39 56

beckyrother:
@shaunanoordin great points. definitely need some feedback when sending script to keyboard – can the button itself change, like [here](https://tympanus.net/Development/CreativeButtons/) – the send icon in the green section, and text appears under it confirming what happens `[current script] is now active in your transcription keyboard`

shaunanoordin:
👍 I like that idea, Becky - simple & clean messaging!

@wgranger I've added existential checks that prevent app crashing when a first-time user accesses and uses the Crib Sheet, but that won't avoid the fact first-time users _still_ won't be able to successfully save any new work until they reload the page. (Yikes, this might affect first-time users of ASM too.)

shaunanoordin:
@wgranger OK the issue with the user preferences seems to have been identified.

When a new user first accesses the project, a new Preference is created for them. The problem is in this code...

```
const fetchPreferences = (user) => {
  return (dispatch) => {
    if (user) {
      user.get('project_preferences', { project_id: config.projectId })
        .then(([preferences]) => {
          if (preferences) {
            dispatch({
              type: FETCH_PREFERENCES,
              preferences
            });
          } else {
            apiClient.type('project_preferences').create({
              links: { project: config.projectId },
              preferences: {}
            //HERE
            //--------------------------------
            }).save((newPreferences) => {
              dispatch({
                type: FETCH_PREFERENCES,
                preferences: newPreferences
              });
            //--------------------------------
            }).catch((err) => {
              console.warn(err.message);
            });
          }
        });
    } else {
     ...
  };
};
```

It appears that the `.save()` call _shouldn't_ a `function` as its argument, as it will result in the API client making a call to the URL `https://panoptes-staging.zooniverse.org/api/project_preferences?function%20(newPreferences)%20{%20%20%20%20%20%20%20%20%20%20%20%20dispatch({%20%20%20%20%20%20%20%20%20%20%20%20%20%20type:%20FETCH_PREFERENCES,%20%20%20%20%20%20%20%20%20%20%20%20%20%20preferences:%20newPreferences%20%20%20%20%20%20%20%20%20%20%20%20});%20%20%20%20%20%20%20%20%20%20}`

I'm thinking instead of...
```
...
}).save((newPreferences) => {
  dispatch({
    type: FETCH_PREFERENCES,
    preferences: newPreferences
  });
}).catch((err) => {
...
```

Shouldn't it be...
```
...
}).save()
.then((newPreferences) => {
  dispatch({
    type: FETCH_PREFERENCES,
    preferences: newPreferences
  });
}).catch((err) => {
...
```
...?

wgranger:
Yep, that above fix seems to work. I tested out the most recent commit with a new user.

PR #40, opened 2018-03-14
Allow Hebrew Text from Keyboard

wgranger:
This branch should cover the needed functionality of typing in Hebrew and clicking buttons in the annotation box to insert Hebrew characters. However, I am curious if this PR is sound from an accessibility standpoint since I am somewhat hijacking the keyboard.


src/components/SelectedAnnotation.jsx
```diff
@@ -60,10 +64,43 @@ class SelectedAnnotation extends React.Component {
...
+    const character = Utility.getHebrewChar(e);
+
+    if (Utility.getKeyCode(e) === KEY_VALUES.Enter) {
+      this.saveText();
+    }
+
+    if (character !== false) {
+      e.preventDefault();
+      this.setState({
+        annotationText: this.inputText.value + character.unicode
```
eatyourgreens: Does this add new text to the end of the existing text, even if the text carat is in the middle of the text in the textarea? For example, if I've clicked on a word so I can correct a spelling mistake.

wgranger: Ahh, good call. This is something I need to straighten out. I will have to put the `<input>` into more of a test drive.

eatyourgreens: After working on the text highlighter tools for PFE, working with the [Selection API](https://developer.mozilla.org/en-US/docs/Web/API/Selection) can get quite complicated too. I wonder if there's a way to let people edit Latin text in the input, just like they would normally, then translate the characters to Hebrew on the fly?

wgranger: What do you mean by translate the characters to Hebrew on the fly? Users are expected to be typing in Hebrew characters directly from the script without the need for translation.

shaunanoordin: I believe what Jim means by "on the fly" is something like this...
```
textInput.onChange = () => {  //Whenever the user types in a new character...
  textInput.text = textInput.text  //...do an "autocorrect" for each letter.
  .replace(/a/ig, HEBREW.ALEF)
  .replace(/b/ig, HEBREW.BETH)
  .replace(/g/ig, HEBREW.GIMMEL)
  ...;
}
```

There are pros and cons to this method, namely:
1. it respects input position (e.g. if the user moves the cursor to the middle of the word and types a character)
2. users can never type or copy-paste English characters. (Wait is this a pro or a con?)
3. it _doooesn't_ (?) interfere with copy-pasting of Hebrew text.
4. it _shooould_ (?) work with native Hebrew keyboards.
5. we don't need to capture keyboard input, which makes it simpler to implement (i.e. no need to have listeners) and avoids us accidentally capturing keyboard shortcuts.

eatyourgreens: Yes. Sorry for the slow reply. I thought it would maybe avoid having to work with browser text selection APIs if you had the volunteers edit Latin text, from a Latin keyboard, and just translated the alphabet to Hebrew when you display the text. I am not sure how feasible this is, but it is easier to implement than a full text editor in JavaScript.

eatyourgreens: I hadn't even thought about trying to detect cut-and-paste events. I believe [React's synthetic event](https://reactjs.org/docs/events.html) supports `onPaste` but I've never actually tried it to see how it works in different browsers.

beckyrother:
shaun++ for tagging me

>Character Input/Translation is captured even if onscreen keyboard is dismissed
If I close the onscreen keyboard, when I press my A key on my physical keyboard, the transcription popup captures that input and inserts the character ש instead of the character a.
This should definitely happen, since the keyboard input takes up so much of the screen.

>Copy Paste...?
>Do we need to consider what happens when a user tries to copy and/or paste text into the field? This is probably a secondary concern, so can be answered later, but is worth thinking about.
Yes, this should be considered.

>Cmd+C and Cmd+V and similar commands are impossible.
>If you try to press Cmd+V to paste some text from (say) your personal hebrew_cribsheet.docx file, you'll instead type in the character ה
this seems bad, is there a solution?


Ok then i read the rest of your message 


It will be confusing if the same keys trigger different characters depending on whether or not the keyboard is open. No matter what we do, users will see an input box and expect conventional input box actions, so we should do our best to match those expectations.

There shouldn't be a need for Latin alphabets at any point – to use modifiers like "unclear", users will Have to use the buttons rather than typing in English. I can't see another way around that. If there's somehow latin text on the page, i guess we tell people to flag it in Talk and move on. 


Does that help at all? Have I made it worse? Happy Friday?

src/components/SelectedAnnotation.jsx
wgranger: Ok. This makes sense and might be the approach I wind up taking. I have started today looking at the Selection API, but it is proving to be a bit tricky. Also, using the `onChange` function rather than `onKeyDown` will help with some of the input changes.

wgranger: This makes sense but I'm having the issue of the cursor moving to the end of the input if the `this.inputText.value` is manipulated with `onChange`, which can be an issue when adding for than one letter in the middle of a word.

shaunanoordin:
@wgranger I've added a [new commit](https://github.com/zooniverse/scribes-of-the-cairo-geniza/pull/40/commits/a17fab49ee739b3b20be564cf9e100891fb8b5a9) that allows us to "whitelist" native Hebrew keyboard input.

This means that if a user types in some actual Hebrew characters using a Hebrew keyboard, there's a check in `Utility.getHebrewChar()` that detects natural Hebrew input and just lets it pass as is.

The rest of `getHebrewChar()` remains the same, so it'll still catch input from, say, an Arabic keyboard and translate that properly to match the onscreen Hebrew keyboard..

src/components/SelectedAnnotation.jsx
```diff
@@ -73,17 +63,22 @@ class SelectedAnnotation extends React.Component {
   }
 
   onKeyDown(e) {
+    if (e.ctrlKey || e.metaKey || e.altKey || e.shiftKey) {
```
wgranger: Not sure what sort of solution we arrived on here, but this works with keeping in copy/paste functionality, etc. It will also allow vowels to be added for users with Hebrew keyboards (although not sure we will necessarily need this).

wgranger:
Ok, this appears to be working much better and I appreciate the comments here and help pairing.

shaunanoordin:
## PR Review (Update)
Nice one, @wgranger ! 👍 OK, the latest changes address a lot of the issues raised from the previous review.
- The Hebrew character insertion (in the transcription popup) works more naturally, respecting the position of the text caret/text selection.
- The Hebrew character insertion also respects users with native Hebrew keyboards.
- Bonus: ESC button now triggers the "cancel transcription" action.
- Cmd+C, Cmd+V works for copy pasting! _(Note to self: check on IE11 when I'm back)_

This update is looking good to go. 👌 Noice.

### Status
LGTM. Merging, go!

PR #41, opened 2018-03-22
Implement SubjectViewer Translations

wgranger:
This PR sets up translations for the app and adds translations for the subject viewer. More translations will need to be added eventually for other content on the page. I decided to use `react-localize-redux` rather than `counterpart` because the former worked better with changing the language state and translations through Redux.

There are a few translations missing for the subject viewer which will need to be added eventually.

### DIFF
--- a/package.json
+++ b/package.json
@@ -25,6 +25,7 @@
   "homepage": "https://github.com/zooniverse/scribes-of-the-cairo-geniza",
   "dependencies": {
     "classnames": "~2.2.5",
+    "counterpart": "~0.18.6",
     "history": "~4.6.3",
     "markdownz": "~7.3.1",
     "panoptes-client": "~2.8.0",
--- a/package.json
+++ b/package.json
@@ -25,7 +25,7 @@
   "homepage": "https://github.com/zooniverse/scribes-of-the-cairo-geniza",
   "dependencies": {
     "classnames": "~2.2.5",
-    "counterpart": "^0.18.1",
+    "counterpart": "~0.18.6",
     "history": "~4.6.3",
     "markdownz": "~7.3.1",
     "panoptes-client": "~2.8.0",
@@ -33,6 +33,7 @@
     "publisssh": "^1.1.0",
     "react": "~15.6.1",
     "react-dom": "~15.6.1",
+    "react-localize-redux": "~2.16.0",
     "react-redux": "~5.0.5",
     "react-rnd": "^5.1.0",
     "react-router": "~4.1.2",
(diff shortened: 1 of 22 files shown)


src/components/SelectedAnnotation.jsx
```diff
@@ -203,7 +205,7 @@ class SelectedAnnotation extends React.Component {
     const notes = emptyText ? 'You cannot save an empty transcription.' : '';
     this.props.dispatch(togglePopup(
       <QuestionPrompt
-        confirm="Yes, delete"
+        confirm={this.props.translate('cribSheet.confirm')}
         deny="No, continue transcribing"
```
eatyourgreens: Should this be localised too?

package.json
```diff
@@ -25,14 +25,15 @@
   "homepage": "https://github.com/zooniverse/scribes-of-the-cairo-geniza",
   "dependencies": {
     "classnames": "~2.2.5",
-    "counterpart": "^0.18.1",
+    "counterpart": "~0.18.6",
```
eatyourgreens: Is counterpart used anywhere? I couldn't see it.

src/components/ActiveCard.jsx
```diff
@@ -26,18 +28,24 @@ class ActiveCard extends React.Component {
     };
   }
 
-  deactivateCard() {
-    this.props.dispatch(activateCard(null));
+  onDelete() {
```
eatyourgreens: Have the new methods in this component crept in from a different PR? They don't seem to be related to translations.

eatyourgreens: nvm. Are these existing methods that have been re-ordered by the linter?

src/components/SelectedAnnotation.jsx
wgranger: We don't have a translation for that yet.

package.json
wgranger: Yes, good call. I forgot to remove this after I decided on `react-localize-redux`.

src/components/ActiveCard.jsx
wgranger: Yep, it's kinda irritating when the linter complains about method orders, but I make the switch to remove the error.

src/components/SelectedAnnotation.jsx
eatyourgreens: I'd localise this in the view here and put placeholder strings in the locale files. No big deal, though, if you're aware that it still needs to be done.

eatyourgreens:
One minor comment about the "No, continue transcribing" label but I'm happy to merge this as-is.

PR #42, opened 2018-03-26
SubmitClassification Modal

wgranger:
This should contain all of the functionality needed to confirm a classification can be submitted. 

The modal is similar to the component on ASM, with a couple exceptions.

1) Functionality does not yet exist to save a classification. Instead, this branch contains a bare bones version of classification submission. I'd like to revisit saved classifications anew in a future PR.

2) You may note some hard-coded text in the `FinishedPrompt` where the question is typed. In ASM, this text is retrieved directly from the workflow resource. In this PR, the text is hardcoded for styling reasons (paragraph break).

### DIFF
--- a/src/components/App.jsx
+++ b/src/components/App.jsx
@@ -3,8 +3,9 @@ import PropTypes from 'prop-types';
 import { connect } from 'react-redux';
 import { Route, Switch } from 'react-router-dom';
 import { ZooHeader, ZooFooter } from 'zooniverse-react-components';
-import { fetchResources } from '../ducks/initialize';
 
+import { fetchResources } from '../ducks/initialize';
+import { generateSessionID } from '../lib/get-session-id';
 import AboutLayout from './about';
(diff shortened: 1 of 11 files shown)


shaunanoordin:
Merge conflict fixed; the review goes on.

Re: Save Progress functionality, no worries, I'm happy to see that feature taken care of later, but what's the plan? I'm more in favour of saving to local storage as opposed to saving to Panoptes, the latter which has more possible points of failure.

Re: I'm not too fussed about the hardcoded text, though I'd like to know if there are plans to Translate() said text.

wgranger:
No, I should take a closer look into those issues you saw in this review. I think this is something that should be taken care of now instead of saving for later. I'll see if I can get a push made shortly.

wgranger:
Ok, this should be good to go now. Turns out `prepareForNewSubject` wasn't reseting the annotations as it should. I'll wait on the UI for classification submission and failure, as this will require a bit of design feedback.

PR #44, opened 2018-04-02
Implement Arabic Keyboard

wgranger:
Most of the code here adds Arabic characters to the project. Implementation to type the characters works similarly to Hebrew characters, but some changes were made to streamline the two. Some changes will need to be made once workflow selection is added to the project so the correct keyboard will be added automatically. 

This keyboard was created based on an [Arabic keyboard layout editor](http://www.keyboard-layout-editor.com/#/gists/ceebf86ce67aabbaf98e8b24acdda6a6).

### DIFF
--- a/src/components/AnnotationKeyboard.jsx
+++ b/src/components/AnnotationKeyboard.jsx
@@ -88,7 +88,14 @@ class AnnotationKeyboard extends React.Component {
       <div className="annotation-keyboard">
         {this.state.lettersByRows.map((row, i) => this.renderRow(row, i))}
         <div>
-          <button className="annotation-keyboard__button space-button">Space</button>
+          <button
+            className={classnames('annotation-keyboard__button space-button', {
+              'char-button__active': this.props.activeKey === 'space'
+            })}
+            onClick={this.props.onLetterClick}
+          >
+            Space
+          </button>
         </div>
       </div>
     );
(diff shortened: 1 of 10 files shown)


wgranger:
In the future, the keyboard language will be selected (Hebrew/Arabic) based on which workflow is chosen from the landing page (Hebrew vs. Arabic translations), so the toggling method we have here is temporary just to test this component.

I'm addressing all other issues in the next push, but will wait on hearing back from font size

beckyrother:
Can we go somewhere in between 14 and 24, maybe like 18? 24 feels crowded in the text box. Other than that everything looks good!

wgranger:
I think the reducers were jumbling up two of the actions with the same name. Both the keyboard and language duck contained an action titled `SET_LANGUAGE`. Making them both unique values seemed to do the trick.

shaunanoordin:
## PR Review (Update)
This is looking good! 👍 The latest changes fix the issues where the Keyboard language and the Website language both shared the same values, which caused shenanigans. So now, trying to transcribe Arabic text while having the website set in Hebrew mode (or etc) is no longer a problem.

### Status
👍 LGTM, let's go.

NOTE: The Arabic/Hebrew debug button still exists and will be taken out when we're ready to properly implement multiple workflows (each workflow having its own unique Keyboard language)

PR #52, opened 2018-04-18
Add Babel's ES6 Rest/Spread operators (i.e. the triple dots: ... )

shaunanoordin:
## PR Overview
This PR updates Babel so we can use the new ES6 Rest and Spread operators, aka the `...` operator.

While this PR doesn't add any new functionality to the website, it does bring the code up to current JavaScript/ES6 functionality, and allows some pretty code writing/duck management. (See Purpose in Scribes)

I feel that enabling `...` should be _standard_ for all our projects due to the minimum amount of work needed to get it running, and the amount of benefits it gives us in organising our code.

### Smooth Operators
Rest operator works like this:
```
function test(first, second, ...theRest) {
  console.log(
    'The first element is ', first,
    ' and the second element is ', second,
    ' and the rest of the arguments, as an ARRAY, are ', theRest
  );
}

test("hello", "world", "i can have", "as many arguments", "as I want", "and they'll all appear", "as an array");
```

Spread operator works like this:
```
var moreAnimals = ['dogs', 'elephants', 'fluffysaurus'];
var animals = ['apes', 'bears', 'cats', ...moreAnimals];  //"spread" each element in moreAnimals into this array

//animals is now: ['apes', 'bears', 'cats', 'dogs', 'elephants', 'fluffysaurus']
```

### Purpose in Scribes
Basically we can reduce...

```
SubjectViewer.defaultProps = {
  annotationInProgress: null,
  annotations: [],
  contrast: false,
  currentSubject: null,
  dispatch: () => {},
  frame: 0,
  imageSize: { width: 0, height: 0 },
  popup: null,
  rotation: 0,
  scaling: 1,
  selectedAnnotation: null,
  showKeyboard: true,
  subjectStatus: '',
  translationX: 0,
  translationY: 0,
  viewerSize: {
    width: 0,
    height: 0
  },
  viewerState: SUBJECTVIEWER_STATE.NAVIGATING
};
```

...to...

```
SubjectViewer.defaultProps = {
  ...ANNOTATIONS_INITIAL_STATE,
  ...SUBJECTVIEWER_INITIAL_STATE,
  ...SELECTED_ANNOTATION_AND_POPUPS_INITIAL_STATE,
  dispatch: () => {},
};
```

### Status
Ready for review. This PR doesn't modify the website functionality, but provides a stepping stone for future dev efforts.

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -2,16 +2,18 @@
   "env": {
     "development": {
       "presets": ["es2015", "react"],
-      "plugins": ["react-hot-loader/babel"]
+      "plugins": ["react-hot-loader/babel", "transform-object-rest-spread"]
     },
     "production": {
       "presets": [
         ["es2015", { "modules": false }],
         "react"
-      ]
+      ],
+      "plugins": ["transform-object-rest-spread"]
     },
     "test": {
-      "presets": ["es2015", "react"]
+      "presets": ["es2015", "react"],
+      "plugins": ["transform-object-rest-spread"]
     }
   }
 }
--- a/package-lock.json
+++ b/package-lock.json
@@ -2,6 +2,7 @@
   "name": "scribes-of-the-cairo-geniza",
   "version": "1.0.0",
   "lockfileVersion": 1,
+  "requires": true,
   "dependencies": {
     "@types/node": {
       "version": "9.4.6",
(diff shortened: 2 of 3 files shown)


.babelrc
```diff
@@ -2,16 +2,18 @@
   "env": {
     "development": {
       "presets": ["es2015", "react"],
-      "plugins": ["react-hot-loader/babel"]
+      "plugins": ["react-hot-loader/babel", "transform-object-rest-spread"]
```
wgranger: Nice. It does seem a bit odd there isn't a "staging" key somewhere here, but I assume somewhere that definition would be moved to "development".

wgranger:
Thanks! This should help clean up this code a good bit.

PR #53, opened 2018-04-19
"Save Progress" feature

shaunanoordin:
## PR Overview
This PR adds the ability for users to save/resume their work in progress, by using the Save Progress button on the Classifier page.
- A `WorkInProgress` duck/function library has been added.
  - For future-proofing I'm reorganising how we code our ducks, starting with WorkInProgress.
- Prerequisite (in this PR): the Workflow and/or Subject duck will need to be updated to better follow a lifecycle Project-Workflow-Subject-Classification dependency lifecycle.
  - This is because when a save project is loaded, it needs to pick a specific workflow, then pick a specific subject, in that order - this may short-circuit the way we currently baked workflow-subject fetch assumptions, if we're not careful.

### Intended design
- When a user visits the Classifier, check if the user has a saved work in progress.
  - If nothing, proceed as normal.
  - If something, prompt user to "start anew" or "continue work in progress" 
    - if start anew, clear all work in progress and load Classifier as normal.
    - if continue, LOAD SAVED WORKFLOW, then LOAD SAVED SUBJECT (in that order!) then create a new empty Classification (with the correct Workflow ID and Subject ID), then load saved annotations.
- When a user submits a Classification SUCCESSFULLY, clear all work in progress before fetching the next Subject.

### Status
The WorkInProgress duck is still a work in progress, and work is progressing.

### DIFF
--- a/src/components/ControlPanel.jsx
+++ b/src/components/ControlPanel.jsx
@@ -8,7 +8,10 @@ import { getTranslate, getActiveLanguage } from 'react-localize-redux';
 
 import { fetchGuide } from '../ducks/field-guide';
 import { toggleDialog, togglePopup } from '../ducks/dialog';
-import { WorkInProgress } from '../ducks/work-in-progress';
+import {
+  WorkInProgress, WORKINPROGRESS_INITIAL_STATE, WORKINPROGRESS_PROPTYPES,
+  getWorkInProgressStateValues
+} from '../ducks/work-in-progress';
 import FieldGuide from '../components/FieldGuide';
 import CribSheet from '../components/CribSheet';
 import TutorialView from '../components/TutorialView';
@@ -194,6 +197,15 @@ class ControlPanel extends React.Component {
           )}
 
           <hr className="white-line" />
+
+          <div>
+            <button className="button">{this.props.translate('infoBox.transcribeReverse')}</button>
+            <button className="button" onClick={()=>{this.props.dispatch(WorkInProgress.save())}}>
+              {this.props.translate('infoBox.saveProgress')}
+            </button>
+            {this.props.wipTimestamp && (<div className="workinprogress-timestamp">{this.props.wipTimestamp.toString()}</div>)}
+            <button className="button button__dark" onClick={this.finishedPrompt}>{this.props.translate('infoBox.finished')}</button>
+          </div>
           
           <div>
             <button className="button" onClick={()=>{
@@ -204,20 +216,11 @@ class ControlPanel extends React.Component {
               this.props.dispatch(WorkInProgress.clear());
             }}>DEBUG CLEAR</button>
             <button className="button" onClick={()=>{
-              console.log('WorkInProgress.save()');
-              this.props.dispatch(WorkInProgress.save());
-            }}>DEBUG SAVE</button>
-            <button className="button" onClick={()=>{
               console.log('WorkInProgress.load()');
               this.props.dispatch(WorkInProgress.load());
             }}>DEBUG LOAD</button>
           </div>
-
-          <div>
-            <button className="button">{this.props.translate('infoBox.transcribeReverse')}</button>
-            <button className="button">{this.props.translate('infoBox.saveProgress')}</button>
-            <button className="button button__dark" onClick={this.finishedPrompt}>{this.props.translate('infoBox.finished')}</button>
-          </div>
+          
         </div>
       </Section>
     );
@@ -268,7 +271,8 @@ ControlPanel.propTypes = {
   }),
   workflow: PropTypes.shape({
     id: PropTypes.string
-  })
+  }),
+  ...WORKINPROGRESS_PROPTYPES
 };
 
 ControlPanel.defaultProps = {
@@ -283,7 +287,8 @@ ControlPanel.defaultProps = {
   tutorial: null,
   tutorialStatus: TUTORIAL_STATUS.IDLE,
   user: null,
-  workflow: null
+  workflow: null,
+  ...WORKINPROGRESS_INITIAL_STATE,
 };
 
 const mapStateToProps = (state) => {
@@ -299,7 +304,8 @@ const mapStateToProps = (state) => {
     tutorial: state.tutorial.data,
     tutorialStatus: state.tutorial.status,
     user: state.login.user,
-    workflow: state.workflow.data
+    workflow: state.workflow.data,
+    ...getWorkInProgressStateValues(state),
   };
 };
 
--- a/src/components/ControlPanel.jsx
+++ b/src/components/ControlPanel.jsx
@@ -8,6 +8,7 @@ import { getTranslate, getActiveLanguage } from 'react-localize-redux';
 
 import { fetchGuide } from '../ducks/field-guide';
 import { toggleDialog, togglePopup } from '../ducks/dialog';
+import { WorkInProgress } from '../ducks/work-in-progress';
 import FieldGuide from '../components/FieldGuide';
 import CribSheet from '../components/CribSheet';
 import TutorialView from '../components/TutorialView';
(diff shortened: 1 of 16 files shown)


shaunanoordin:
## PR Update
This PR now has functional save/resume Work In Progress features.
- Save/Load Work In Progress is currently available via debug buttons on the Classifier page.
- ❗️ IMPORTANT: the Workflow, Subject, etc ducks have been updated to better respect the data lifecycle.
  - For example, loading a new Workflow now resets Subject, Classification, and Annotations.
  - Loading a new Subject resets Classification and Annotations.
- ❗️ IMPORTANT: the Workflow, Subject, etc ducks that use `apiClient` now properly `return` the API client. See below:
```
//DERP we had an error in our ducks all along! Because of this, this.props.dispatch(fetchSubject1()); couldn't be chained, because the dispatch returns undefined!
const fetchSubject1 = (subjectId = null) => {
  ...
  apiClient.type('subjects').get(subjectId)
  ...
}

//THIS version will work, because now this.props.dispatch(fetchSubject2()); returns a Promise. RETURN ALWAYS
const fetchSubject2 = (subjectId = null) => {
  ...
  return apiClient.type('subjects').get(subjectId)
  ...
}
```

### Functional Features
- Users can save WIP data. (Using a debug button)
- Users can load WIP data. (Using a debug button)
- Saved WIP data is deleted either when:
  - A user successfully loads WIP data.
  - A user successfully  submits (all queued) classifications

### Features To Be Built
- Proper UI/UX needed:
  - Save Progress buttons needs functionality.
  - Load Progress modal/popup should appear when the user loads the Classifier page.

❗️ The work in this PR sets the stage for users to select/change Workflows, so be aware of all the fundamental framework being set.

### Status
WIP

shaunanoordin:
## PR Update
The UI for Save and Resume Progress have been implemented properly. Please see notes above for intended step-through functionality.

Tagging @beckyrother and @snblickhan for visual design and text advisory. NOTE: text has not been set for translatability (let's pretend that's a valid word).

screen shot 2018-04-20 at 19 49 56
_When Save Progress is clicked, the "Last save" timestamp appears._

screen shot 2018-04-20 at 19 50 16
_When the Classifier page is loaded when the user has saved work in progress, this "Resume Progress?" modal appears. Note that the modal blocks all possible interactions until it's been responded to._

### Status
This PR is now ready for review; tagging @wgranger

beckyrother:
Thanks Shaun! Here's my design notes
First image:
- Please use the serif font, left aligned, for the save information. It can match the style of the Attribution information.
- Please move the save info to under the Finished button to avoid making that button move around if the information appears or changes.

Second image:
- Please use the same style as modals like [this one](https://invis.io/7RH9A5FCUND#/265052990_Transcribe-7-Done-_No). 
- Can the buttons be the same width as the ones in the subject info toolbar?

src/components/ControlPanel.jsx
```diff
@@ -46,11 +52,21 @@ class ControlPanel extends React.Component {
...
+    if (nextProps.tutorial !== this.props.tutorial && !WorkInProgress.check(this.props.user)) {
       Tutorial.checkIfCompleted(nextProps.tutorial, nextProps.user, nextProps.preferences).then((completed) => {
         if (!completed) { this.toggleTutorial(); }
       });
     }
+    
+    //Now check if the user has any work in progress.
+    console.log('+++ componentWillReceiveProps', this.props.user !== nextProps.user, WorkInProgress.check(nextProps.user));
+    if (this.props.user !== nextProps.user && WorkInProgress.check(nextProps.user)) {
+      console.log('+++ WorkInProgressPopup', nextProps.user);
```
wgranger: Were these logs simply for debugging?

src/components/WorkInProgressPopup.jsx
```diff
@@ -0,0 +1,59 @@
...
+          <div className="work-in-progress-popup__message">
+            <span className="work-in-progress-popup__header">Resume Work In Progress?</span>
+            <p>
+              We detected that you have some work in progress saved.
+              You can continue your saved work, or start with a new page.
+            </p>
+          </div>
+          <div>
+            <button className="button" onClick={this.startNewWork}>New Page</button>
+            <button className="button button__dark" onClick={this.resumeWorkInProgress}>Resume Work</button>
```
wgranger: We will need to request translations for these items once this is merged.

src/ducks/subject.js
```diff
@@ -54,57 +67,118 @@ const subjectReducer = (state = initialState, action) => {
...
-        currentSubject,
-        id: currentSubject.id,
-        queue: getState().subject.queue,
-        type: FETCH_SUBJECT_SUCCESS,
-        favorite: currentSubject.favorite || false
-      });
+      
+      const subjectQuery = { workflow_id };
+
+      // TODO What if the fetched queue is empty?
```
wgranger: Are you referring to error handling here? I think this would only occur if there weren't any subjects associated with in workflow.

src/components/ControlPanel.jsx
wgranger: It doesn't look like the `WorkInProgressPopup` is triggered on this line in the instance that the landing page is visited first and then the user navigates to "transcribe" on the nav bar. I think that is because the user never changes (`this.props.user !== nextProps.user`) since it has already been retrieved from the initial load.

wgranger:
Overall, this looks pretty clean and nice compared to the work we've done on ASM. Also, thanks for organizing other aspects of the code and returning some of those api calls that might've been lost. 

I think this code will continue to evolve a bit as we fine tune some aspects on the site, namely workflow selection. 

There are a couple minor comments here, and just one main blocking comment concerning how the `WorkInProgressPopup` might fail to appear if the user initially loads the landing page.

src/components/ControlPanel.jsx
shaunanoordin: Argh! Thanks, these `console.log()s` need to go.

And thank you for that catch - I only tested that the LoadProgress prompt works if I refresh on the Classifier page.

Possible solution: I should put a workInProgress check on _componentDidMount..._ but ONLY if user has already logged in. (Otherwise, logged-in users who refresh the page on the Classifier will accidentally trigger the anonymous user's Save Progress data.) I realise I'm kinda giving the shaft to the non-logged-in users who want to save/retrieve their progress... but then, logically, they shouldn't be able to see the Save Progress button anyway. Hmm.

src/ducks/subject.js
shaunanoordin: Yup! This wouldn't happen in Scribes of the Cairo Geniza (unless something really effs up) but this comment came part of the "generalised CFE library" code, and I wanted to make sure that some day we'll be able to account for the situation when the app tries to get a queue from 1. a project with no subject sets, 2. a project with subjects sets that have no subjects, 3. a project with pretty much fully classified subjects OK there's nothing to do what do you do now.  The _generalised_ fetchSubject() feature shouldn't just return an empty array, it should also raise a flag of some sort to indicate, "Hey, the fetch was successful... but we just have nothing, yo."

shaunanoordin:
Thanks for the review, @wgranger ! Here's on my fix list:
- Becky's design updates
- Get rid of console.log()s
- Bugfix: A LOGGED IN user should be able to retrieve their saved data if they visit the Classifier page either directly, or navigated there from the home page/about page/etc. (non-logged-in users don't get anything, eh)

👍 BRB, gotta get coding!

shaunanoordin:
Update update: TRANSLATIONS!

We now have translations for the Work In Progress prompt, woo woo.

src/ducks/work-in-progress.js
```diff
@@ -0,0 +1,268 @@
...
+      
+      //First, load the Workflow, then the Subject, then the Annotations.
+      //Finally, clear all saved WIP data.
+      dispatch(fetchWorkflow(workflowId)).then(() => {
+        return dispatch(fetchSubject(subjectId));
+      }).then(() => {
+        return dispatch(loadAnnotations(annotations));
+      }).then(() => {
+        console.log('WorkInProgress.load() success');
+        return dispatch(clear());
```
wgranger: Maybe I missed this in an earlier review, but why is it necessary to clear the `localStorage` here? If a user loads their saved annotation and then closes their browser, their work will be lost. Instead of being cleared, can the work remain and just be updated with any following save? 

I

wgranger:
New commits look great! I only made a comment as it doesn't seem intuitive that your saved work would be removed from `localStorage` once loaded (and only saved back into `localStorage` once you explicitly hit save again). In my opinion saved annotations should remain saved locally unless you explicitly ask for a new subject or successfully submit those annotations.

shaunanoordin:
@wgranger Good point - I've updated the code so that when a user Resumes Work In Progress, their saved data _isn't_ cleared out. WIP data is _only_ deleted when A. the user decides "naw, I'll start a new manuscript" or B. the user successfully submits their Classification.

I remember now why the previous ASM code deleted local storage on load, and why it doesn't matter to Scribes: ASM had _multiple_ save systems, and we decided that the manual Panoptes save should _overrule_ the emergency localStorage save. This is why localStorage was _purposely made volatile_ - so it avoided any logical conflicts between the manual save.

Since we only have _one_ save system for Scribes, there's no reason to make localStorage volatile.

Nice catch 👍

wgranger:
Great! Looks good and is more succinct than the ASM code. Merging.

PR #54, opened 2018-04-20
Use Pandora and Update Translations

### LINKED ISSUE #50: Complete Remaining Translations, also Pandora
This issue is really twofold:

- [ ] The site still needs some minor translations, and documentation for this is being kept [on a Google Doc](https://docs.google.com/document/d/1VtvgYKeHX0WkJ52QBSzgX7flcFgwiGtjqi7B07atezA/edit).
- [ ] Pandora needs to be set up in the app so translations can be made and viewed directly via the Pandora frontend.

wgranger:
So, this PR does a good bit, and I probably should have broken it up into smaller PRs, but I got on a roll, and I apologize. Note the large diff from the package.lock, and I can amend this somehow if needed.

The main items tackled here are:
- Retrieve resource translations from Pandora for the `TutorialView` and `FieldGuide` components.
- Organize styled components into a "styled" folder to more closely reflect our approach on PFE.
- Update translations of content throughout the site.

Closes #50


shaunanoordin:
Merge conflicts detected after merging PR #44 (Implement Arabic Keyboard) so let me resolve those first.

shaunanoordin:
## Quick thoughts
Functionality looks good 👍 
- I can browse pages, make Annotations, view the Tutorial, view the Field Guide, and submit Classifications just fine as I constantly switch between English/Hebrew/Arabic.
- LtR/RtL orientation of the website still correctly reflects the selected website language.
- Some parts are still missing translations - e.g. the Submit Classification prompt - but this is expected.
- I can't tell if anything's missing from the Pandora implementation, and since it _looks_ like everything is in place, I'm going to give that an 👌 
- I'll run through the code now to see that I didn't miss anything.

src/ducks/translations.js
```diff
@@ -0,0 +1,96 @@
...
+const loadTranslations = (translated_type, translated_id) => {
+  return (dispatch, getState) => {
+    const languages = ['ar', 'he', 'en'];
+    dispatch({
+      type: FETCH_TRANSLATIONS,
+      translated_type,
+      translated_id
+    });
+    languages.map((language) => {
+      apiClient.type('translations')
```
shaunanoordin: When we call an asynchronous `dispatch(fetchWhatever())` that uses `apiClient`, we want to make sure it returns a Promise. i.e. we want to make sure we write `return apiClient.type('whatever')...`

In this case, I'm not sure what loadTranslations is returning, if it's returning anything. `undefined` maybe?

The reason I need to check this is because a few lines below this, we have...
```
const loadResources = () => {
  return (dispatch) => {
    Promise.all([
      dispatch(loadTranslations('field_guide', '40')),
      dispatch(loadTranslations('tutorial', '207'))
    ]).then(() => {
      dispatch({ type: FETCH_TRANSLATIONS_SUCCESS });
    }).catch(() => {
      dispatch({ type: FETCH_TRANSLATIONS_ERROR });
    });
  };
};
```
...and IF `dispatch(loadTranslations())` returns `undefined`, then `Promise.all` will cause `FETCH_TRANSLATIONS_SUCCESS` to trigger BEFORE all the `loadTranslation()` calls succeed.

I'm going to test this and see whattup.

shaunanoordin: BOO! Promises are a bit of a headache, as usual. I modded to code to look like this...

```
const loadTranslations = (translated_type, translated_id) => {
  return (dispatch, getState) => {
    //bla bla bla
    languages.map((language) => {
      apiClient.type('translations')
        .get({ translated_type, translated_id, language })
        .then(([translations]) => {
          if (translations) {
            //bla bla bla
            console.log('+++ GOT IT FOR ', language);
            dispatch({
              type: SET_STRINGS,
              payload: {
                [translated_type]: resource
              }
            });
          }
        });
    });
  };
};

const loadResources = () => {
  return (dispatch) => {
    Promise.all([
      dispatch(loadTranslations('field_guide', '40')),
      dispatch(loadTranslations('tutorial', '207'))
    ]).then(() => {
      console.log('+++ SUCCESS');
      dispatch({ type: FETCH_TRANSLATIONS_SUCCESS });
    }).catch(() => {
      console.log('+++ FAILURE');
      dispatch({ type: FETCH_TRANSLATIONS_ERROR });
    });
  };
};
```

...and sure enough, console.log looks like this:
```
+++ SUCCESS
+++ GOT IT FOR  en
+++ GOT IT FOR  he
+++ GOT IT FOR  ar
+++ GOT IT FOR  ar
+++ GOT IT FOR  he
+++ GOT IT FOR  en
```

OK, bad news and good news. 

Bad news, the current way that this PR does fetch translations doesn't work, because `FETCH_TRANSLATIONS_SUCCESS` triggers _before_ all translations are fully fetched.

Good news... this _may_ not matter, if we've been REALLY robust with our existential checks for `reduxStore.translations.strings` - i.e. we check that `reduxStore.translations.strings.['english']` exists before we render `reduxStore.translations.strings.['english'].burpingSounds`

shaunanoordin:
## Not So Quick Thoughts: Translations Promises = 😢 

Blargh! OK, so Translations that get text data from Pandora use apiClient + Promises, but the way we structure our Promises aren't good.

Bad: In the `translations` duck, `Promise.all()` doesn't work properly with `loadTranslations()`, so the `Promise.all()` actually thinks it's done its job (flags a success) before the multiple `loadTranslations()` finish theirs.

Good: it may not matter, as long as our existential checks for the translated string data is robust.

Bad: our existential checks for the translated string data is a bit spotty?

Good: well, at least we only need to worry about FieldGuide and TutorialView, as only those two components use translation data from Pandora. (Other components use locally stored translation files, so there's no need to worry about the async fetches)

### FieldGuide

```
FieldGuide.findTranslations(language) {
  const items = this.props.guide.items.slice();

  if (this.props.translationStatus === TRANSLATION_STATUS.READY) {
    const translated = this.props.translatedGuide[language];
    items.map((item, i) => {
      item.title = translated[`items.${i}.title`];
      item.content = translated[`items.${i}.content`];
    });
  }
  return items;
}
```

`this.props.translationStatus` checks that the Translation status is ready before doing anything, but as established before, Promise.all() incorrectly marks the status as ready before every bit of data is successfully fetched. 🤷‍♂️ 

Possible point of improvement:
```
const translated = this.props.translatedGuide[language];
```

This line assumes too much about what exists and what doesn't. A simple check can make this much safer, perhaps something like...

```
const translated = (this.props.translatedGuide && this.props.translatedGuide[language])
  ? this.props.translatedGuide[language]
  : [];  //What's the safest "default value" to use here? Empty array? Empty object?
```

### TutorialView

The fix for TutorialView should reflect the fix for FieldGuide, perhaps...
```
render() {
  //...bla bla bla...
  const translation = this.props.translatedTutorial[language];  //...changing this line...
```
...to...
```
const translation = (this.props.translatedTutorial && this.props.translatedTutorial[language])
  ? this.props.translatedTutorial[language]
  : {};  //Again, I'm not sure what the proper default value should be. Empty array? Empty object?
```

shaunanoordin:
## Promise.all + Array.map combinations

Some thoughts: if we _really_ wanted to make the multiple-Promise-translation-fetch thing work (NOTE: we probably _don't need to_ for this Beta launch!), we could do something like this...

```
const loadTranslations = (translated_type, translated_id) => {
  return (dispatch, getState) => {
    const languages = ['ar', 'he', 'en'];
    dispatch({
      type: FETCH_TRANSLATIONS,
      translated_type,
      translated_id
    });
    Promise.all(
      languages.map((language) => {
        return apiClient.type('translations')
          .get({ translated_type, translated_id, language })
          .then(([translations]) => {
            if (translations) {
              const resource = getState().translations.strings[translated_type];
              resource[language] = translations.strings;
              console.log('+++ GOT IT FOR ', language);
              dispatch({
                type: SET_STRINGS,
                payload: {
                  [translated_type]: resource
                }
              });
            }
            return language;
          });
      })
    ).then((meepmorp) => {
      //For funsies, the value of `meepmorp` should be `['en', 'ar', 'he']`, but we won't actually be using this.

      return 'YAY';
    });
  };
};

const loadResources = () => {
  return (dispatch) => {
    Promise.all([
      dispatch(loadTranslations('field_guide', '40')),
      dispatch(loadTranslations('tutorial', '207'))
    ]).then((beepboop) => {
      //For funsies, the value of `beepboop` should be `['YAY', 'YAY']`, but we won't actually be using this.
      
      console.log('+++ SUCCESS');
      dispatch({ type: FETCH_TRANSLATIONS_SUCCESS });
    }).catch(() => {
      console.log('+++ FAILURE');
      dispatch({ type: FETCH_TRANSLATIONS_ERROR });
    });
  };
};

```

I haven't tested this out yet, so take my thoughts with a grain of salt.

src/ducks/translations.js
wgranger: Ok, this makes sense. I recall seeing some questionable behaviors in writing this code initially that supports your assumptions above. This is something I'll have to nail out in a future commit based on your review of this PR.

wgranger:
Changes have been made per request. With this newest commit, the translations fallback on English if they have not been returned from the translations endpoint. I tried editing the `loadTranslations` function a bit to work correctly with the promises, but it still needs a bit of tweaking. I'm going to save this for after beta though, since it isn't imperative this is working perfectly for beta launch.

shaunanoordin:
👍 Nice! This PR now looks great, and I just need to solve some conflicts before I merge. BRB

PR #56, opened 2018-04-20
Add Text Modifiers

### LINKED ISSUE #49: Setup Text Modifiers
For the most part, the structure of these text modifiers will be similar to the setup on ASM, and a prototype can be [found on InVision](https://projects.invisionapp.com/d/main#/console/10495701/270077124/preview). Text modifiers are as follows:

*Text modifiers — Final Version*
_Arabic:_
Insertion <‏إضافة>
Deletion {حذف}
Damaged [عطل]
Drawing #‏رسمة#
Grid       #شبكة#

_Hebrew:_
Insertion <נוסף>
Deletion {מחוק}
Damaged [פגום]
Drawing #ציור#
Grid      #גריד#
Divine Name <’ה>

wgranger:
Closes #49 

This PR sets up text modifiers for the app. A couple notes:

1) The regex code might look strange because it is reading the text from right to left. Everything is all turned around with the text direction. We will have to edit this if we decide to change how the open and closed brackets are written.

2) We have double "divineName" and "divine" translations because I am still waiting on the translation for the former. The difference between the two, in concept, is that "divineName" would translate to "Divine Name" (etc. for all languages) for the `SelectedAnnotation` UI and "divine" would handle the word that gets put into the actual textbox.


wgranger:
I made a comment in Slack about the preference for tags as we have them written ltr or `[insertion/]Hello[insertion]`. We're going to proceed with ltr for beta but be prepared to switch if the research group identifies a preference towards a flipped alignment.

Issue #48, opened 2018-04-17
Setup Page Reverse Button


wgranger:
Some UI elements are still up for debate on this, namely what sort of indicator is shown to the volunteer when viewing the front or back. Also, some sample data is needed to have tests of this working on staging.

wgranger:
Closed with #57

PR #57, opened 2018-04-25
Frame Switching - Allow Switching Between Front Page and Reverse Page

shaunanoordin:
## PR Overview
This PR adds functionality to the "Transcribe Page Reverse" button on the Classifier Page's toolbar. As it says on the tin, it flips the page shows the other side, allowing the users a whole new image to annotate. 
- This assumes that Subject in Scribes of the Cairo Geniza has EXACTLY two images, corresponding to the front page and the back page. (Each image = a "frame", in code terms)
- As a result, each annotation/transcription submitted in the Classification will have a `frame` value, which indicates which page the annotation/transcription was made on.
- We're using a 0-index, so frame 0 is for first front page, and frame 1 is for the reverse page.

### Dev Notes

Surprisingly, the groundwork for frame switching is already well baked into the project code, and not much else is needed. Classifications _and_ annotations already respect the `frame` value, and the Subject Viewer seems to render the correct page nicely (with annotations for the front page being separate from annotations for the reverse page) - this PR mostly seems to involve activating the "switchThePagePlease()" function in the "page-switching" button.

Some testing still needs to be done though, as I'm seeing blank images for the reverse pages. This _might_ be due to the Subjects on staging not having a reverse image.

### Status
WIP

### DIFF
--- a/src/components/ControlPanel.jsx
+++ b/src/components/ControlPanel.jsx
@@ -7,6 +7,7 @@ import { getTranslate, getActiveLanguage } from 'react-localize-redux';
 
 import { fetchGuide } from '../ducks/field-guide';
 import { toggleDialog, togglePopup } from '../ducks/dialog';
+import { changeFrame } from '../ducks/subject-viewer';
 import {
   WorkInProgress, WORKINPROGRESS_INITIAL_STATE, WORKINPROGRESS_PROPTYPES,
   getWorkInProgressStateValues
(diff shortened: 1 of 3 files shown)


shaunanoordin:
## PR Update
Update added - a fix has been added for the bug where the frame didn't reset when the user submitted a Classification.

Also, bonus, testing has confirmed that the "flip the page" functionality is working - I've updated the staging version of Scribes of the Cairo Geniza with Subjects that have two faces (you'll recognise these test images veeery quickly) and I'm able to annotate both sides of the manuscript.

### Status
OK, everything's running - this PR's ready for review 👍

src/components/ControlPanel.jsx
```diff
@@ -208,7 +209,10 @@ class ControlPanel extends React.Component {
           <hr className="white-line" />
 
           <div>
-            <button className="button">{this.props.translate('infoBox.transcribeReverse')}</button>
+            {(this.props.frame === 0)
+              ? <button className="button" onClick={()=>{this.props.dispatch(changeFrame(1))}}>{this.props.translate('infoBox.transcribeReverse')}</button>
+              : <button className="button" onClick={()=>{this.props.dispatch(changeFrame(0))}}>{this.props.translate('infoBox.transcribeFront')}</button>
```
wgranger: Can we hide this button if there is not a reverse side? (`currentSubject.locations.length === 1`, or something like that)

shaunanoordin: Can do, gimme a second. 👌

shaunanoordin:
## PR Update
The "Reverse Page" button now only shows if the Subject has two page images (or more).

wgranger:
Closes #48

PR #64, opened 2018-04-27
Introduce Workflow Selection

wgranger:
This PR introduces workflow selection in two places: buttons on the landing page and also a dropdown menu on the project header.

### DIFF
--- a/src/components/Home.jsx
+++ b/src/components/Home.jsx
@@ -3,6 +3,7 @@ import { ZooniverseLogo, ZooniverseLogotype } from 'zooniverse-react-components'
 import { connect } from 'react-redux';
 import { getTranslate, getActiveLanguage } from 'react-localize-redux';
 import PropTypes from 'prop-types';
+import { withRouter } from 'react-router-dom';
 import Penn from '../images/penn.png';
 import JTSLogo from '../images/jts_logo.png';
 import Princeton from '../images/princeton_geniza.png';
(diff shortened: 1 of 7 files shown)


eatyourgreens:
If it’s any help, PFE stores your workflow choice in preferences, so that you always get the last workflow you worked on when you load a classify page.

wgranger:
I can make these changes pretty quickly. I think the best solution for now (without devoting too much time to a new UI component this close to beta, is to load the most recent workflow choice from the preferences. 

Also, we could assume simple Hebrew is the default choice, but I worry that would lead to that workflow retiring subjects much quicker than the other workflows. I can load a random workflow on first visit if that is preferable.

shaunanoordin:
## Update
Following IMLS meeting discussions, we've decided that we'll apply the "default workflow update/fix" in a separate PR. For now (i.e. the period leading up to Beta), we are OK to have "Easy Hebrew" as the default workflow. The preferred long-term solution is the workflow popup/prompt.

This PR will now be merged so we can properly configure the app to function on the production environment. 👍

wgranger:
It looks like the mechanism to set your most recent workflow choice in preferences is built into the PFE frontend, so if a user selects a workflow in the CFE, it will not be updated in the preferences. 

This would require some more rigging in this project to set up the preferences update. Because of that, I might opt to devoting any dev time on this to go ahead and set up the selection prompt.

PR #65, opened 2018-04-27
Fix Small Issues

wgranger:
This branch was initially created to check cross-browser compatibility, but the branch also takes care of small errors on the site. For the most part, the site looks good on Chrome, Safari, Firefox, and Edge (with minor display issues on Edge I'm not worrying about until after beta).


wgranger:
Let me give this branch another run through, and rebase, before we merge. I want to see how this looks with recent changes to master (concerning rtl). Sam and I had a short talk about switching keyboards, and I think it will only be available in the Hebrew workflows (as that is the only place where there will be text from both languages). I'll make a change in another commit.

wgranger:
Updated. I think I got all the changes correct in the about page, but it's worth double checking.

shaunanoordin:
## PR Review (Update)
Just checked this PR, and everything still looks good, as far as I saw.

Merging, go!

PR #68, opened 2018-05-01
Workflow select prompt

wgranger:
This PR adds a workflow select prompt when no workflow is present.

This PR also attempts to resolve existence issues with workflows. The organizational changes here should take care of issues with workflow changes and retrieving new tutorials. 

Please also note any changes that may need to be made with styling.

### DIFF
--- a/src/components/ProjectHeader.jsx
+++ b/src/components/ProjectHeader.jsx
@@ -16,12 +16,17 @@ class ProjectHeader extends React.Component {
     this.toggleDropdown = this.toggleDropdown.bind(this);
   }
 
+  componentWillReceiveProps(next) {
+    if (this.props.location.pathname !== next.location.pathname && this.props.showWorkflows) {
+      this.toggleDropdown(false);
+    }
+  }
+
   changeLanguage(language) {
     this.props.dispatch(setLanguage(language));
   }
 
-  toggleDropdown() {
-    const nextState = !this.props.showWorkflows;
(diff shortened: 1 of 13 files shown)


shaunanoordin:
Conflict detected, let's resolve this.

shaunanoordin:
Conflict resolved. Testing resumes.

src/components/ProjectHeader.jsx
```diff
@@ -16,12 +16,17 @@ class ProjectHeader extends React.Component {
     this.toggleDropdown = this.toggleDropdown.bind(this);
   }
 
+  componentWillReceiveProps(next) {
+    if (this.props.location.pathname !== next.location.pathname && this.props.showWorkflows) {
```
shaunanoordin: I like that we have a bit of code here that closes the Classify->Workflow Selection dropdown when we change pages, though we should get into a better practice of noting down what's happening with a short comment.

src/components/WorkflowPrompt.jsx
```diff
@@ -0,0 +1,18 @@
...
+
+const WorkflowPrompt = () => {
+  return (
+    <div className="workflow-prompt">
+      <div className="workflow-prompt-test">
+        <div className="workflow-prompt__header">
+          <span>Choose a Workflow</span>
+          <hr className="plum-line" />
+        </div>
+        <WorkflowDropdown className="workflow-prompt-popup" />
```
shaunanoordin: Ooh, I like that we're recycling the component here!

Issue #83, opened 2018-05-09
Translation Edits & Additions


snblickhan:
- [x] Add English name, "Scribes of the Cairo Geniza", next to the Hebrew name in the homepage

- [x] Reformat dropdown menu in Hebrew -- text is being presented outside the box (see below) 
![example](http://i67.tinypic.com/fc7fbd.png)

- [x] Switch text for show/collapse fragment details (as per image below) 
![example](http://i63.tinypic.com/qyua9z.png)

- [x] Update text for "Show tutorial" to הדרכה למשימת התעתוק (as per image above)

- [x] update Hebrew translations for all workflow names (list below)

1. "Easy Hebrew":    עברית קלה
2. "Challenging Hebrew":     עברית – דרגת קושי גבוהה
3. "Easy Arabic":    ערבית קלה 
4. "Challenging Arabic":    ערבית – דרגת קושי גבוהה 

TRANSCRIPTION BOX:
- [x] update Hebrew translations for text modifiers (list below)
1. "Insertion":    תוספת
2. "Deletion":     מחיקה
3. "Grid":   טבלה/שרטוט
4. "Divine Name":    שם האל

- [x] "Show modern characters":      הצג מקלדת רגילה

- [x] Translations needed for pop-up box when user tries to close an unsaved transcription

1. "Close annotation":   סגור תעתוק
2. "Are you sure you want to close without saving?":    ?בטוחים שאתם רוצים לסגור בלי לשמור 
3. "Yes, close without saving":    כן, סגור בלי לשמור
4. "No, continue transcribing":    לא, המשך לתעתק

- [x] Punctuation is incorrectly placed in the instructions on the transcription box. Example below is Hebrew, but it's the same problem in Arabic. ![example](http://i67.tinypic.com/10o0eio.png)

snblickhan:
Looks like everything here is done. There will be a new Translation issue coming soon, so I'll close this one to avoid confusion, unless someone says otherwise.

PR #85, opened 2018-05-10
Update Hebrew Translations

wgranger:
This PR updates Hebrew translations as identified in #83. One issue of that PR still remains, and some Arabic translations are still missing.

### DIFF
--- a/src/components/Home.jsx
+++ b/src/components/Home.jsx
@@ -50,7 +50,7 @@ const Home = ({ currentLanguage, dispatch, history, language, rtl, translate }) 
           <ZooniverseLogotype width="100px" />
           <h1>{translate('topNav.site')}</h1>
           {language !== LANGUAGES.ENGLISH && (
-            <h1>Scribes of the Cairo Geniza</h1>
+            <h2 className="h2-font">Scribes of the Cairo Geniza</h2>
           )}
         </div>
         <hr className="plum-line" />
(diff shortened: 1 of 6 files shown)


src/components/SelectedAnnotation.jsx
```diff
@@ -192,12 +192,12 @@ class SelectedAnnotation extends React.Component {
   closePrompt() {
     this.props.dispatch(togglePopup(
       <QuestionPrompt
-        confirm="Yes, close without saving"
-        deny="No, continue transcribing"
+        confirm={this.props.translate('closeAnnotation.confirm')}
```
eatyourgreens: Maybe `const { translate } = this.props;` at the top of this function, to save some repetition in these props.

src/components/Home.jsx
```diff
@@ -49,6 +49,9 @@ const Home = ({ currentLanguage, dispatch, history, rtl, translate }) => {
         <div>
           <ZooniverseLogotype width="100px" />
           <h1>{translate('topNav.site')}</h1>
+          {language !== LANGUAGES.ENGLISH && (
+            <h1>Scribes of the Cairo Geniza</h1>
```
eatyourgreens: I'm not sure what's happening here. Does this add a second `h1` if the language isn't English? Documents should really only have one top-level heading.

wgranger: The research team wanted the English variant kept when switching to Arabic and Hebrew. I can change it to `h2`, as that would probably make more sense.

src/components/SelectedAnnotation.jsx
wgranger: Ah, good call. This will help readability.

PR #87, opened 2018-05-11
Finish Stats Section

wgranger:
This PR finishes the stats section on the homepage. Technically, it should be ready to merge, but it is marked RFC because I'd like to optimize one aspect of the code and would like a suggestion:

1) When the stats section loads, all workflows are retrieved to correctly calculate phase two stats, which makes it unnecessary to ever fetch another workflow once these are retrieved.
2) However, the `fetchWorkflow()` function returns a promise. If we instead said in that function "instead of an api call/promise, just return that workflow, if you already have it," the `fetchWorkflow()` calls in the app break because they expect a promise return to chain the `.then()` call when `fetchSubject()` is next called.
3) Any suggestions? This code works just fine without the optimization, but it would save a few API calls if we found a solution.

### DIFF
--- a/src/components/HomeStatistics.jsx
+++ b/src/components/HomeStatistics.jsx
@@ -1,42 +1,65 @@
 import React from 'react';
+import { connect } from 'react-redux';
+import PropTypes from 'prop-types';
 import Connect from './Connect';
 
-export default function HomeStatistics() {
-  return (
-    <div className="statistics">
-      <div className="statistics__stat-content">
-        <div className="statistics__pie-chart">
-          <svg viewBox="0 0 200 200">
-            <circle cx="100" cy="100" r="100" fill="#EBDFD6" stroke="#472B36" strokeWidth="5px" />
-            <path d="M100 100 L 100,0 A100,100 0 0,1 200,100 z" fill="#472B36" />
-          </svg>
-          <span>Percent Complete</span>
-          <hr className="plum-line" />
-        </div>
-        <div className="statistics__statistics">
-          <h2>Statistics from phase two of Scribes of the Cairo Geniza</h2>
-          <div className="statistics__stat">
-            <div>
-              <span>831</span>
-              <span>Volunteers</span>
-              <hr className="plum-line" />
-            </div>
-            <div>
-              <span>12,473</span>
-              <span>Classifications</span>
-              <hr className="plum-line" />
-            </div>
-            <div>
-              <span>3,147</span>
-              <span>Completed Subjects</span>
-              <hr className="plum-line" />
-            </div>
+class HomeStatistics extends React.Component {
+  render() {
+    const project = this.props.project;
+    if (!project) return null;
+
+    return (
+      <div className="statistics">
+        <div className="statistics__stat-content">
+          <div className="statistics__pie-chart">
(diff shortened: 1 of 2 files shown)


src/ducks/workflow.js
```diff
@@ -98,9 +102,40 @@ const toggleSelection = (show) => {
   };
 };
 
+const fetchPhaseTwoWorkflows = () => {
+  return (dispatch) => {
+    const c = config;
```
srallen: This seems unnecessary to assign to a variable.

src/ducks/workflow.js
```diff
@@ -98,9 +102,40 @@ const toggleSelection = (show) => {
...
+    const allWorkflows = {
```
srallen: This being an object doesn't really seem useful to you since in the component you convert it back to an array anyway. I would start with an array of workflow ids to begin with to pass into the request directly:
```
apiClient.type('workflows').get(arrayOfWorkflowIds).then(allWorkflows => dispatch({ type: FETCH_ALL_WORKFLOWS, allWorkflows})).catch(err => console.error(err))
```

Then in your component, you can iterate over an array directly since the response will be an array of workflows.

src/components/HomeStatistics.jsx
```diff
@@ -1,42 +1,105 @@
...
+
+  render() {
+    const { project, allWorkflows } = this.props;
+    if (!project || !allWorkflows) return null;
+    let classificationsCount = 0;
+    let percentComplete = 0;
+    let retiredSubjects = 0;
+    let subjectsCount = 0;
+
+    Object.keys(allWorkflows).map((id) => {
```
srallen: If `allWorkflows` is an array to begin with you avoid converting back to an array by calling `Object.keys()`. `forEach` would be more appropriate to use here since you don't need to return an array.

src/components/HomeStatistics.jsx
```diff
@@ -1,42 +1,105 @@
...
+      const workflow = allWorkflows[id];
+      if (workflow) {
+        classificationsCount += workflow.classifications_count;
+        retiredSubjects += workflow.retired_set_member_subjects_count;
+        subjectsCount += workflow.subjects_count;
```
srallen: I assume this here is happening because the workflows are on that project being shared with workflow stats you don't want to show here. Another good reason to break them apart into two projects under a shared organization so you can take advantage of all the stats math the API does for you already serialized on the project resource.

src/ducks/workflow.js
wgranger: I want to keep `allWorkflows` as an object since the user could be swapping between workflows throughout the project. Storing all workflows as an object seemed an easier way to switch to a different workflow if the object used workflow ids as object keys.

src/components/HomeStatistics.jsx
wgranger: `allWorkflows` is actually an object. It is kept an object so it is easier to retrieve a different workflow (English, Arabic, Hebrew) throughout the app by using literal notation with the workflow id.

shaunanoordin:
Alright, this PR is getting merged, but first I want to thank @srallen for helping review the code early on. Now let's get our stats on!

PR #89, opened 2018-05-14
Handle Language Issues

wgranger:
Two issues are addressed in this PR:

1) The scroll image on the home page has been replaced when in rtl orientation. The original image was simply flipped, which led to the text not making sense.
2) The `alefLam` character has been changed a bit. Originally the character was only displaying one character (I believe an alef) when the character is actually the combo of the two. To account for this, the character has to try in two characters in succession (alef and lam), which will also need to advance the pointer position two spaces since the character is actually two unicode characters (see code).


wgranger:
@simensta That is to be expected, since the alef lam character is actually two characters combined. I don't believe it's standard for an Arabic keyboard to have an alef lam button, so we're having to hack it a bit here by adding two characters in succession.

Issue #92, opened 2018-05-15
Full Launch Checklist


snblickhan:
HOME PAGE
- [x] Need to grey out all workflows for pre-launch -- Penn team have PR starting on June 11, and they would like the site to be accessible (via URLs in those articles) but not functional (all buttons greyed-out for the time being). This will then be removed before full launch. (see below for screenshot of links that need to be disabled) #99 ![example](http://i64.tinypic.com/5vrqco.png) ![example](http://i65.tinypic.com/1z3135d.png)
- [x] Update "TRANSCRIBE" dropdown -- remove full list of options & reduce selections to "Continue Saved Work" or "Start a New Transcription". The latter option opens a list of workflows, including Classification & Keyword Search (similar to the boxes available on the Home page). #84 
- [x] Typo in the "What is Zooniverse?" paragraph: "discoeries" is missing the letter V (should read "discoveries") #98 
- [x] Remove duplicate logo for Cambridge ![example](http://i63.tinypic.com/n3u5ns.png) #98 
- [x] Replace current Princeton logo ("Princeton Geniza Project") with logo at 
https://drive.google.com/file/d/1rS7I3PAxvAA4Bx9SoTmyUqWWf3mZ_Ayu/view #98
- [x] Replace current "e-Lijah lab" logo with **two** Haifa logos (one for e-Lijah lab, one for Genizah Research Centre): https://drive.google.com/drive/folders/1G-L1_8S2Q47yirRBAzOPgYa2vevLkXBN #98
- [x] New Home Page intro text: https://docs.google.com/document/d/15iTcVfi7UKiPGI0MQxTQy_kC2bvFW6kvbTbkzdyeAzo/edit?usp=sharing #98 
- [x] New Special Thanks text: https://docs.google.com/document/d/1Jvff6XdZiOySPsOJ1FNC72t16k7Ow3re8h1Y0hsYHyM/edit?usp=sharing #98 

TRANSCRIPTION INTERFACE
- [x] When you click on an existing transcription (to revise your work), you can't press "Done" until you've changed something. This is a problem if a user wants to review a transcription, but changes their mind & doesn't want to edit the text -- they can't press Done. #84 
- [x] Crib sheet list appearing off the bottom edge of the modal #94  ![example](http://i67.tinypic.com/105xop2.png)
- [x] Add external link icon to "Library Catalog Page" link in Subject Info menu #98 

TRANSCRIPTION MODAL
[...]

PR #98, opened 2018-06-07
Address Post-Beta Issues

wgranger:
This PR handles many of the issues brought up in #92. Items covered in this PR are denoted in that issue.


src/components/about/content-provenance-ar.jsx
```diff
@@ -20,7 +16,7 @@ function Content() {
...
-          <a href="http://openn.library.upenn.edu/html/genizah_contents.html" target="_blank">http://openn.library.upenn.edu/html/genizah_contents.html</a></p>
+          <a href="http://openn.library.upenn.edu/html/genizah_contents.html" rel="noopener noreferrer" target="_blank">http://openn.library.upenn.edu/html/genizah_contents.html</a></p>
```
shaunanoordin: I'm just reading up on the importance of `rel="noopener"` and `rel="noreferrer"`; I was hitherto unaware of the significant performance impact and the security issue. TIL.

src/ducks/subject.js
```diff
@@ -182,7 +182,9 @@ const prepareForNewSubject = (subject) => {
     dispatch(createClassification(subject));
     dispatch(changeFrame(0));
     if (aggregationWorkflows.indexOf(activeWorkflowID) >= 0) {
-      subject && dispatch(fetchAggregations(subject.id, activeWorkflowID));
+      const aggWorkflowId = activeWorkflowID === config.challengingHebrew ?
+        config.hebrewKeyword : config.arabicKeyword;
+      subject && dispatch(fetchAggregations(subject.id, aggWorkflowId));
     } else {
       dispatch(clearAggregations());
```
shaunanoordin: I just noticed that `fetchAggregations()` doesn't properly reset the `redux.aggregations` store like `clearAggregations()` does - specifically, when fetching a new aggregation, the `aggregations.showHints` doesn't change (which is fine), and `aggregations.keywordWorkflow` retains its old value until we enter a success state. (Which... should be fine.)

This might cause some weird edge case shenanigans, but I don't believe there's anything to worry about, in practice. Still, worth noting it here in case future me is looking for where past me dropped the ball.

src/components/about/content-provenance-ar.jsx
shaunanoordin: Now that I see the problem, I also see that there is a _small truck load_ of `<a target="_blank">` links that still don't have `rel="noopener noreferrer"` as a safety. This is another item on the future ToDo.

src/components/Dialog.jsx
```diff
@@ -35,7 +35,7 @@ class Dialog extends React.Component {
   }
 
   render() {
-    const zInd = this.props.component === this.props.focusedDialog ? '100' : '0';
+    const zInd = this.props.component === this.props.focusedDialog ? '100' : '5';
```
shaunanoordin: I don't see the difference between z-index 0 and z-index 5 for a Dialog that doesn't have focus; I see that both values work more or less the same (i.e. the "non-focused" dialog appears behind the dialog with focus) given the scenarios I'm testing on OSX+Chrome. Can you tell me what I'm missing, or what this was intended to fix?

(Are there browsers where z-index 0 meant the non-focused dialog disappears into the page behind it?)

PR #100, opened 2018-06-11
Hide Workflows from WorkflowSelection Component

wgranger:
With #84 merged, the `WorkflowSelection` component needed to be updated to reflect the work that was done in #99. This PR also deletes the `WorkflowDropdown` component that is no longer being used.


shaunanoordin:
Also the "Coming soon" sign for Easy Arabic is a little off (the span is inside the button) so I'll fix that too.

shaunanoordin:
## PR Update
This PR has now been updated so the newly disabled links now have their links disabled. This tautological statement has been brought to you by the Department of Redundancy Department, which brought you this statement that's tautological.

### Status
Merging and deploying.

PR #107, opened 2018-06-14
Improve Transcription popup (SelectedAnnotation component) position

shaunanoordin:
## PR Overview
This PR improves the position of the transcription popup (i.e. the SelectedAnnotation component).

When the transcription popup _pops up,_ (i.e. after creating a new Annotation or clicking on an existing one) the position of the popup should NOT obscure the annotation marks, and should appear below (vertically beneath) them.

screen shot 2018-06-14 at 18 29 43

_The transcription popup appearing vertically below the annotation marks. This allows readers to read the line of text they just marked._

### Status
WIP

Currently:
- `<Dialog>` component now accepts a "default x-y offset" as an optional argument. This allows us to position the Dialog (hence, the SelectedAnnotation component) in a way that's _not_ centred on the window.
- `SubjectViewer.figureOutActualXYOfAPointInTheImage()` calculates the actual x-y coordinates (i.e. relative to the window) of a point on the image (i.e. the point has coordinates relative to the rotated/scaled/panned image)

Next:
- Now that we know the x-y coordinates of the start/end points of an annotation, we need to calculate a decent y-offset for the transcription popup window.

### DIFF
--- a/src/components/WorkflowSelection.jsx
+++ b/src/components/WorkflowSelection.jsx
@@ -198,7 +198,7 @@ WorkflowSelection.defaultProps = {
 const mapStateToProps = state => {
   const user = state.login.user;
   const userHasWorkInProgress = user && WorkInProgress.check(user);
-
+  
   return {
     //Does the user currently have a page being actively annotated, (e.g. user
     //navigated away from the Classifier page and wants to return), or have
@@ -206,7 +206,9 @@ const mapStateToProps = state => {
     //We need to know if the user has any work that can be retrieved (either
     //from the Redux store of local storage) so we can prompt them to continue.
     activeAnnotationExists: (!!state.workflow.data && !!state.subject.currentSubject) || userHasWorkInProgress,
-    adminMode: state.login.adminMode,
+    adminMode: (!/^http:\/\/localhost:3000\//i.test(window.location.href))  //TEMP
+      ? state.login.adminMode
+      : true,
     currentLanguage: getActiveLanguage(state.locale).code,
     translate: getTranslate(state.locale),
     user: state.login.user,  //Needed, otherwise component won't update when it detects a user login.
--- a/src/components/WorkflowSelection.jsx
+++ b/src/components/WorkflowSelection.jsx
@@ -206,7 +206,7 @@ const mapStateToProps = state => {
     //We need to know if the user has any work that can be retrieved (either
     //from the Redux store of local storage) so we can prompt them to continue.
     activeAnnotationExists: (!!state.workflow.data && !!state.subject.currentSubject) || userHasWorkInProgress,
-    adminMode: false,
+    adminMode: state.login.adminMode,
     currentLanguage: getActiveLanguage(state.locale).code,
     translate: getTranslate(state.locale),
     user: state.login.user,  //Needed, otherwise component won't update when it detects a user login.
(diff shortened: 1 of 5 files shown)


shaunanoordin:
## PR Update
Alright, this PR is ready - I've added a `SubjectViewer.guesstimateBestOffsetForAnnotationBox()` that figures out the best y-position to place the annotation popup.

Note that the function had some sanity checks to prevent the annotation popup from appearing too far below (or too far above, as unlikely as they might be) the window viewport, but it turns out that it causes conflicts with the browser's (well, Chrome's) natural self-correcting mechanism where the window would scroll down anyway to show the annotation popup. I've left the faulty(?) "correct code" in this PR, in comments, in case you wanted to see what's up there.

### Status
@wgranger this is now ready for review 👌

src/components/WorkflowSelection.jsx
```diff
@@ -198,15 +198,15 @@ WorkflowSelection.defaultProps = {
...
+  
   return {
     //Does the user currently have a page being actively annotated, (e.g. user
     //navigated away from the Classifier page and wants to return), or have
     //saved work in progress? (e.g. user reloaded the website after a crash)
     //We need to know if the user has any work that can be retrieved (either
     //from the Redux store of local storage) so we can prompt them to continue.
     activeAnnotationExists: (!!state.workflow.data && !!state.subject.currentSubject) || userHasWorkInProgress,
-    adminMode: state.login.adminMode,
+    adminMode: false,
```
wgranger: Is there a reason this was changed to false?

wgranger:
Since the transcription box will now open lower (under the most recent annotation dot placed) should we open the transcription pane with the keyboard closed as the default? This is probably a UX question for @beckyrother

wgranger:
Overall this code looks pretty good. The transcription pane did open over some dots when the document was rotated, but I don't think this is a huge issue as the box still moves out of the way some. There is one item that I think was accidentally changed (see inline comment) and I think we should also wait back to hear if the keyboard should be closed on open before a merge. Overall, good stuff!

src/components/WorkflowSelection.jsx
shaunanoordin: D'OH!!!

shaunanoordin:
D'OH! @wgranger thank you so much for spotting that derp I made with `adminMode: false` in WorkflowSelection.jsx.

I previously set that mapping to `adminMode: true` for my localhost development, (because otherwise I'd have to login every time I refreshed the page). However, when I reverted that temporary change, I incorrectly "reverted" it to `adminMode: false` instead of the correct `adminMode: state.login.adminMode` mapping.

## PR Update
The correct fix has been applied with the latest commit, and yup, good call - let's wait to hear if the keyboard should be open/closed initially. @beckyrother, hit us with that UI/UX goodness!

beckyrother:
@wgranger @shaunanoordin I think the keyboard should be open by default – assuming most people working on this project don't have a Hebrew or Arabic physical keyboard, they need a way to type. I know the modal is big, and if there's another layout I'm all ears, but unfortunately this is the best I could come up with.
We should be able to remember a person's choice between lines, though, if they close the keyboard.

shaunanoordin:
Thanks Becky! 👌 Alright, we'll keep the keyboard open by default. @wgranger, is there anything else this PR is missing?

wgranger:
Nope! This PR should be good to go.

PR #110, opened 2018-06-19
Submit Keyboard Information with Hebrew Manuscripts

### LINKED ISSUE #96: Record keyboard script used during transcription
Note: for Hebrew workflows only.

The research team would like to know which script type (the options that can be set as the keyboard: Ashkenazi Square, etc.) is being used in the keyboard for each individual transcription.

There are a few options for this:
1. Record what keyboard was selected _at the time the entire classification was completed_ (i.e. the last keyboard selection used before submitting).
2. Record which keyboard was used _for each line_ of a transcription (this could be helpful for documents featuring multiple script types, but may not be necessary if it's going to be a lot of work).
3. If possible, also record whether keyboard was open or closed (to see % of volunteers who are typing from their own keyboard).

wgranger:
Closes #96 

This PR submits the type of keyboard script and a boolean if the keyboard UI was open with each classification sent on a Hebrew manuscript. This feature should be deployed by full launch.

### DIFF
--- a/src/ducks/classification.js
+++ b/src/ducks/classification.js
@@ -5,6 +5,7 @@ import { fetchSubject } from './subject';
 import { WorkInProgress } from './work-in-progress';
 import { resetView } from './subject-viewer';
 import { getSessionID } from '../lib/get-session-id';
+import { LANGUAGES } from './languages';
 
 const CLASSIFICATIONS_QUEUE_NAME = 'classificationsQueue';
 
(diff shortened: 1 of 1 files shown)


wgranger:
Yes, the second point is a concern and I should include modern keyboard as a recordable option. I'll push up a commit now.

shaunanoordin:
## PR Review (Update)
Nice, this is looking good! Merging, go!

PR #118, opened 2018-06-20
Use Standard Hebrew and Arabic Fonts

### LINKED ISSUE #115: Check Hebrew & Arabic fonts
All Hebrew text should be [Rubik](https://fonts.google.com/specimen/Rubik).

All Arabic text should be [Amiri](https://fonts.google.com/specimen/Amiri).

Once fonts are changed, we may need to up the size for Hebrew & Arabic, but we'll do that check after the font update.

wgranger:
Closes #115 

This was the only solution I could think of to close this issue, and I'll admit, it isn't optimal, but I'm not really sure how to address it. 

@beckyrother please ping me at some point to chat about some potential unintended consequences of resolving this issue. It would be easier to show on this branch rather than explain, but basically, many of your font sizes (and in turn, element dimensions) will change with a site-wide font change.

### DIFF
--- a/src/index.tpl.html
+++ b/src/index.tpl.html
@@ -9,7 +9,7 @@
       Genizah Research Unit at Cambridge University Library share Cairo Geniza fragments
       with the Zooniverse community for the first time!">
     <meta name="viewport" content="width=device-width, initial-scale=1">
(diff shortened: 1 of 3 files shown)


beckyrother:
Just to keep everyone in the loop, I'm talking to an Arabic speaker on Friday morning to see about the size of the Arabic fonts. I'll post an update after to let everyone know what she says.

PR #124, opened 2018-06-21
Add Timeout safety

### LINKED ISSUE #39: setTimeout() in <ScriptReferences /> might cause leak
## Bug needs fixing
The Script References feature has a `.sendToKeyboard()` function that utilises `setTimeout()`.

While great UI design, the issue is that the `setTimeout()`'s function could possibly try to use `this.setState()` when the `this` in question (the Script Reference component) has been unmounted - this leads to potential memory leak issues.

- See this code: https://github.com/zooniverse/scribes-of-the-cairo-geniza/blob/d99d5b7479f6edbd0962ad41e8f093c367744f89/src/components/ScriptReferences.jsx#L27-L37
- This follows analysis made in PR #38 - see that for full details

### Status
Bug is memory-related and should be addressed when possible.

shaunanoordin:
## PR Overview
This PR fixes #39 , which states that there's a potential for bad things to happen if a Timeout function attempts to use parts of React component that's not mounted (notably, this.setState()).

Technically simple, but there were shenanigans in the implementation.

### Dev Notes
1. Uh... I had to name the variables `this._isMounted` instead of `this.isMounted` because the latter actually causes the app to crash. Not kidding, this is what Chrome complained about: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Errors/Getter_only

Does anyone have insights into this? Does the `_` underscore prefix somehow make variables private vars, via Babel or React? Something something? This is one of those implicit things that I was hitherto unaware of.

2. I only checked `if (this._isMounted)` and not `if (this && this._isMounted)` because the `()=>{}` statement should ensure a `this` reference is always bounded (bound? binded?).

3. Technically, only ScriptReferences.jsx had any real potential to cause trouble since it's the only component whose Timeout function uses `this.setState()`, whereas the other components only trigger Redux store updates. Still, functionally, it's better this way to avoid weird "something happened on a page that I wasn't looking at" issues.

### Status
@wgranger behold the weirdness.

### DIFF
--- a/src/components/ScriptReferences.jsx
+++ b/src/components/ScriptReferences.jsx
@@ -28,11 +28,11 @@ class ScriptReferences extends React.Component {
   }
   
   componentDidMount() { 
-    this.isMounted = true;
+    this._isMounted = true;
   }
 
   componentWillUnmount() {
-     this.isMounted = false;
+     this._isMounted = false;
   }
 
   sendToKeyboard() {
@@ -43,7 +43,7 @@ class ScriptReferences extends React.Component {
     this.props.dispatch(toggleKeyboard(true));
     this.setState({ keyboardSent: true });
     setTimeout(() => {
-      if (this.isMounted) {
+      if (this._isMounted) {
         this.setState({ keyboardSent: false });
       }
     }, 4000);
--- a/src/components/ScriptReferences.jsx
+++ b/src/components/ScriptReferences.jsx
@@ -26,14 +26,6 @@ class ScriptReferences extends React.Component {
       keyboardSent: false
     };
   }
-  
-  componentDidMount() { 
-    this._isMounted = true;
-  }
-
-  componentWillUnmount() {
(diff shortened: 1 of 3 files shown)


eatyourgreens:
Best practice in React is that you should cancel any pending timeouts in `componentWillUnmount`. `setTimeout()` returns an ID that you can store and use later as an argument to `removeTimeout`.

shaunanoordin:
That's a way better idea - OK, I'll redo this PR with a clearTimeout() on
the component unmount event instead.

srallen:
`isMounted` used to be a part of the React API. It's been deprecated and here's why: https://reactjs.org/blog/2015/12/16/ismounted-antipattern.html

Using an underscore in javascript is just convention to indicate a private class method or property. The best solution I think is what @eatyourgreens suggests.

wgranger:
Nice! Good succinct fix.

PR #134, opened 2018-10-03
Update Hebrew Translations

wgranger:
This PR updates Hebrew translations. It also adds Arabic translations for the landing page paragraph about the project.


package.json
```diff
@@ -15,7 +15,7 @@
   },
   "engines": {
     "node": "^8.1.2",
-    "npm": "^5.0.3"
+    "npm": "^6.0.0"
```
shaunanoordin: Why are we using npm 6?

eatyourgreens: Automatic security audits?

shaunanoordin: That makes sense as to why, though my local configuration (managed with `nvm`) is throwing up its arms in confusion: my `node 8` uses `npm 5.5.1`, but changing it to `npm 6+` end up with me using `node 10`, which is equally incompatible. 🤷‍♂️ 

That said, this is just a quirk of my local setup that's making it a faff to run, not really a problem with the code or PR itself. I'm going to re-read my `nvm` documentation.

eatyourgreens: You can run `npm install -g npm` to upgrade npm without changing node version. I still find it weird that you use `npm` to install npm.

PR #138, opened 2018-10-24
Remove Feedback from Top Nav

### LINKED ISSUE #137: Remove feedback tab from CFE
Now that we're ready for the full launch, we need to remove the "Feedback" tab from the CFE.

wgranger:
Closes #137 

Just removing the feedback link from the top nav. I kept the feedback translations in case we need the in the future.


shaunanoordin:
I'll deploy this shortly.

PR #143, opened 2019-02-05
Gray Consensus Dots

wgranger:
The branch is called `add-gray-lines`, but we're opting solely for gray dots here. This PR should handle all the work needed to display consensus dots. This PR mainly duplicates functionality from ASM, with slight modifications. Functionality included here:

- Account for all reductions in staging and production (would be good to double check there are no graphQL errors in testing).
- Handles previous consensus dots changing with frame change. The "Easy Hebrew" workflow on staging has a subject set with a back (second) page.
- Adds messaging on hover over consensus dot.

* Also, the package-lock was updated here with a _ton_ of changes, but I can get rid of this if requested.


shaunanoordin:
## Investigation Complete!

DERP! I spent way too long on this. 🤦‍♂️

### Diagnosis

In regards to Issue 1, I just realised why `consensusReached` is _always false:_

```
//In src/ducks/consensus-line.js
//--------
const constructLines = (reductions, frame) => {
  ...
        const data = {
          points,
          frame,
          consensusReached:
            annotation.consensus_score >= CONSENSUS_SCORE_TO_RETIRE ||
            annotation.number_views >= MINIMUM_VIEW_TO_RETIRE
        };
  ...
};
//--------

//In src/config.js
//--------
const CONSENSUS_SCORE_TO_RETIRE = 3;
const MINIMUM_VIEW_TO_RETIRE = 7;

...

export { env, config };
//--------
```

... `CONSENSUS_SCORE_TO_RETIRE` and `MINIMUM_VIEW_TO_RETIRE` **aren't exported by config.js**  so it ends up as `undefined` in `constructLines()`

### Change Request

```
export { env, config, CONSENSUS_SCORE_TO_RETIRE, MINIMUM_VIEW_TO_RETIRE };
```

...should fix the problem pretty quickly. 👍 To test that the consensus lines are working for yourself, drop the Consensus Score and Min Views to 1.

wgranger:
Sorry! That's my bad I forgot the export. I'll go ahead and make that change now. I'll go ahead and see what the dots look like with a stroke, which should help the issue of visibility.

wgranger:
Stroke definitely helps visibility. 

screen shot 2019-02-12 at 10 34 40 am

PR #144, opened 2019-02-05
Use Pandora Translations for Finished Prompt

wgranger:
We currently use local translations for the `FinishedPrompt` component. However, in order to keep content and translations for this modal flexible for project owners, the `FinishedPrompt` now renders text from Pandora. 

If there is no translation from Pandora, the modal will fallback on local translations (header and instructions) or content in the English translation (answer labels).

This is how text will render from the project builder to the modal:
screen shot 2019-02-05 at 2 11 32 pm

### DIFF
--- a/src/components/FinishedPrompt.jsx
+++ b/src/components/FinishedPrompt.jsx
@@ -62,15 +62,22 @@ class FinishedPrompt extends React.Component {
       )
         .filter(task => isAQuestionTask(task, this.props.workflow))
       : [];
+    const language = this.props.currentLanguage;
+    const translations = (this.props.translatedWorkflow && this.props.translatedWorkflow[language])
+      ? this.props.translatedWorkflow[language] : null;
+    const currentTask = tasks && tasks.length && tasks[0].taskId;
+    const instructionsTranslation = translations && translations[`tasks.${currentTask}.help`];
+    const headerTranslation = translations && translations[`tasks.${currentTask}.question`];
+
+    const instructions = instructionsTranslation || translate('finished.instructions') + translate('finished.whenReady');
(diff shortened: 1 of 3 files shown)


src/components/FinishedPrompt.jsx
```diff
@@ -62,15 +62,22 @@ class FinishedPrompt extends React.Component {
...
         .filter(task => isAQuestionTask(task, this.props.workflow))
       : [];
+    const language = this.props.currentLanguage;
+    const translations = (this.props.translatedWorkflow && this.props.translatedWorkflow[language])
+      ? this.props.translatedWorkflow[language] : null;
+    const currentTask = tasks && tasks.length && tasks[0].taskId;
+    const instructionsTranslation = translations && translations[`tasks.${currentTask}.help`];
+    const headerTranslation = translations && translations[`tasks.${currentTask}.question`];
+
+    const instructions = instructionsTranslation || translate('finished.instructions') + translate('finished.whenReady');
```
shaunanoordin: [MINOR] If we fall back to local translations here, `translate('finished.instructions') + translate('finished.whenReady')` will probably look pretty weird because it'll lack any spacing between the two sentences.

src/components/FinishedPrompt.jsx
```diff
@@ -62,15 +62,22 @@ class FinishedPrompt extends React.Component {
...
+    const header = headerTranslation || translate('finished.allTranscribed');
```
shaunanoordin: [MINOR] This makes perfect sense to me, but I'm wondering how easy it'll be for future devs to understand that we're using Pandora translations and then falling back to local translations.

PR #145, opened 2019-02-05
Use Different "Copy" Icon for SVG Overlay

wgranger:
With the update to FontAwesome 5, our previous solution to display an FA icon in an SVG `<text>` element no longer works. I've removed the Font Awesome className here in lieu of a more straightforward Unicode character.

You should see this icon, and the SVG overlay, if you enter the "Challenging Hebrew" workflow in staging.

### DIFF
--- a/src/components/AggregationsPane.jsx
+++ b/src/components/AggregationsPane.jsx
@@ -124,12 +124,11 @@ class AggregationsPane extends React.Component {
               </text>
               <text
                 id={`${uniqueKey}_icon`}
-                fontFamily="FontAwesome"
                 fontSize="2.5em"
                 x={x + BUFFER}
                 y={y - BUFFER}
               >
-                &#xf0c5;
+                &#x2398;
               </text>
             </g>
           );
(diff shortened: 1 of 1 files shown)


shaunanoordin:
Ah! Changing the icon to `&x2398;` isn't necessary, we just need to change the font name because FontAwesome 5's font name is... `Font Awesome\ 5 Free`. Uh, with that slash, apparently.

```
<text
  id={`${uniqueKey}_icon`}
  fontFamily="Font Awesome\ 5 Free"  <!--Previously this was just "Font Awesome"-->
  fontSize="2.5em"
  x={x + BUFFER}
  y={y - BUFFER}
>
  &#xf0c5;
</text>
```

Must be the way that it's parsing spaces and numbers.

wgranger:
I'll give it a go. I thought I tried this, but it could be I forgot the word "Free." Looks like there's a conflict I'll have to take care of either way.

wgranger:
Good eye. You should be good to go now.

shaunanoordin:
(Font)Awesome, let's get this PR merged! 👍

Issue #61, opened 2018-04-26
View Site in All Browsers


wgranger:
...and possibly in mobile if only for the landing and about pages.

PR #148, opened 2019-02-11
Fix tutorial

### LINKED ISSUE #147: RTL Edge Tutorial Error
**How to replicate:** 
- When in Edge/IE, change the site text to Arabic or Hebrew. 
- Open the tutorial. 
- Advance the tutorial.
- You should see a problem with the way each tutorial step overlaps each other as the tutorial advances.

image

wgranger:
Fixes #147 

This PR removes the `StepThrough` component in the tutorial component in favor of a more simple approach (similar to the PFE solution). 

The first commit here also runs an `npm audit fix` to update outdated packages.

Please keep a close eye on how the tutorial now renders in IE/Edge.


shaunanoordin:
Merge conflicts? Nooo!

src/components/TutorialView.jsx
```diff
@@ -96,10 +89,8 @@ class TutorialView extends React.Component {
   }
 
   isFinalStep() {
-    const currentStep = (this.stepThrough && this.stepThrough.state && this.stepThrough.state.step)
-      ? this.stepThrough.state.step + 1 : 0;
-    const maxSteps = (this.stepThrough && this.stepThrough.props && this.stepThrough.props.children)
-      ? React.Children.count(this.stepThrough.props.children) : 0;
+    const currentStep = this.state.stepIndex;
+    const maxSteps = this.props.tutorial.steps.length - 1 || 0;
```
shaunanoordin: I'm not sure what the point of this logic is; if `this.props.tutorial.steps.length` is 0, then `maxSteps` will be `-1`, because in JS, `-1 || 0` will result in `-1`. Did you mean `Math.max(this.props.tutorial.steps.length, 0)` ?

shaunanoordin: Ah wait, I get what's happening here - you're worried that `this.props.tutoria.steps.length` might be undefined. If that's the case, we should check for the existence of `this.props.tutoria.steps`, since `.length` should be a given if `steps` is an Array

src/components/TutorialView.jsx
```diff
@@ -108,13 +99,26 @@ class TutorialView extends React.Component {
...
+    this.setState({
+      stepIndex: index
+    });
+  }
+
   render() {
     if (!this.state.loaded) {
       return null;
     }
+    const totalSteps = this.props.tutorial.steps.length || 0;
```
shaunanoordin: Echoing my comments from earlier, if we're worried that `this.props.tutorial.steps.length` might be `undefined`, then it'd be best if we checked the existence of `this.props.tutorial.steps`. Otherwise, this line of code doesn't do much; if `steps` exists, then it should have `.length` of 0 or more (it's unlikely to be a non-Array), and if `step` _doesn't_ exist, then this line of code will crash anyhow.

wgranger:
Closes #61 

This should be the last PR needed to get the app ready for Chrome/Safari/Edge.

src/components/TutorialView.jsx
wgranger: This logic needs to be cleaned up generally. Yes, that `||` was mainly being used as an existential check.

wgranger:
Conflicts have been fixed. I had to run another `npm install`, so the packages-lock is different again.

PR #173, opened 2019-06-13
Update ZRC and run `npm audit fix`

shaunanoordin:
## PR Overview

The intent of this PR is to update a number of dependencies in the project, for security purposes. This mirrors zooniverse/anti-slavery-manuscripts#291 and zooniverse/edu-api-front-end#155, with a special note of caution given ASM's penchant for the website to _look fine_ when running locally but crashing when deployed.

- Zooniverse dependencies have been updated
- `npm audit fix` was run; most notable is the updated `css-loader` which necessitated...
- ...changes to the webpack config files.

### Status

Compiles fine on localhost (+OSX+Chrome), but... you saw what happened with ASM.

Weirdly enough, to continue this test, I need to see if this works on a live web server, so I'm first going to _deploy the project_ (since we don't have staging, alas) before merging. That way, if this effs up, it's easier to revert back to a `master` version that I know works.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -4,6 +4,21 @@
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
+    "@babel/runtime": {
+      "version": "7.4.5",
+      "resolved": "https://registry.npmjs.org/@babel/runtime/-/runtime-7.4.5.tgz",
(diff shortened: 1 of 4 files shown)


shaunanoordin:
### PR Update

Tests look good. 👍 `https://scribesofthecairogeniza.org` has been updated with no issues, so that just means that ASM is the odd duck.

Merging, go.

PR #202, opened 2020-12-21
Remove publisssh in favor of s3-batch-uploads

srallen:
Gets this app consistent with our other apps being updated to use s3-batch-uploads. I don't see an npm script for deploying to staging, so I don't see a good way to test this outside of the dry run flag, which has been added to the upload config script. This can be removed when ready.

### DIFF
--- a/bin/upload.js
+++ b/bin/upload.js
@@ -1,0 +1,30 @@
+/*
+  An S3 uploader with some useful default settings for Zooniverse static files.
+  Accepts an S3 path, relative to zooniverse-static, as its only argument.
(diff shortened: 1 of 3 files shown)


shaunanoordin (approved):
## PR Review

This PR changes our deployment tool, from publisssssh to s3-batch upload.

Dry run looks good. While it ain't great that we don't have a staging for scribes to test this on, I'm pretty confident this will work on production. I've got an upcoming change that needs to be manually deployed anyway so we'll see the new upload in action then. (And if it fails for whatever reason, I can always revert to publissssssssh)

LGTM, removing the dryRun flag, and merging. 👌

PR #208, opened 2021-03-26
Add Github Actions. Remove Travis

srallen:
It looks like there's no staging? @shaunanoordin is that right?


shaunanoordin (approved):
## PR Review

This PR removes Travis, and adds an automated (production) deployment action. LGTM. 👍 

@srallen That's correct, there's no staging _website_ for Scribes - staging _data_ is only used for localhost development. 👌

PR #240, opened 2022-04-28
Use shared workflows for deploys

eatyourgreens:
- use `zooniverse/ci-cd` workflows to run the production deploy.
- run a test build on changes to any PR branch, to catch any breaking changes to build tools.


eatyourgreens:
This should fix production deploys, which are currently broken.

